Issue #52, opened 2022-01-19
Uploading reasults requires approved association


aristizabal95:
Current state of the project allows users to upload results to a benchmark they haven't yet been approved for. Even though there's no obvious concern (security-wise) for that, it could be confusing and lead to users thinking their results will be used even though they haven't been approved for that. 

Given this, we should raise an error from the server if a user tries to POST to /results unless the specified entities related to that result are all officially part of the benchmark.

Additionally, this implies some changes from the CLI testing script, as we need to approve the association request before running the result submission command.

PR #74, opened 2022-02-17
Added validation for Result submission

### LINKED ISSUE #45: [Question] Generating benchmark execution results on pending association
Right now, a dataset owner can create and upload results to the server before an association has been approved. Should this be the case? Results from pending or rejected associations should not be accesible by the benchmark owner, but that doesn't mean that a data owner eager on running models on his machine should wait for an approval from the benchmark. If anything, his efforts are simply not taken into account, but I don't see any problem on allowing results creation before approved association.

Any thoughts?

johnugeorge:
Results are accepted only for
1.  Approved benchmark-dataset association
2. Approved benchmark-MlCube association

Fixes: #52 
Fixes: #115 

Closes #45

### DIFF
--- a/cli/cli.sh
+++ b/cli/cli.sh
@@ -69,9 +69,9 @@ echo "====================================="
 echo "Approving dataset association"
 echo "====================================="
 # Log in as the benchmark owner
-BENCHMARK_OWNER_TOKEN=$(curl -s -X POST "http://127.0.0.1:8000/auth-token/" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{  \"username\": \"testbenchmarkowner\",  \"password\": \"test\"}" | jq -r '.token')
+BENCHMARK_OWNER_TOKEN=$(curl -s -X POST "$SERVER_URL/auth-token/" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{  \"username\": \"testbenchmarkowner\",  \"password\": \"test\"}" | jq -r '.token')
 # Mark dataset-benchmark association as approved
-curl -s -X PUT "http://127.0.0.1:8000/datasets/1/benchmarks/1/" -H  "accept: application/json" -H  "Authorization: Token $BENCHMARK_OWNER_TOKEN" -H  "Content-Type: application/json" -d "{  \"approval_status\": \"APPROVED\"}"
+curl -s -X PUT "$SERVER_URL/datasets/1/benchmarks/1/" -H  "accept: application/json" -H  "Authorization: Token $BENCHMARK_OWNER_TOKEN" -H  "Content-Type: application/json" -d "{  \"approval_status\": \"APPROVED\"}"
 echo "====================================="
 echo "Running benchmark execution step"
 echo "====================================="


server/result/serializers.py
```diff
@@ -7,3 +10,27 @@ class Meta:
...
+    def validate(self, data):
+        benchmark = data["benchmark"]
+        mlcube = data["model"]
+        dataset = data["dataset"]
+        last_benchmarkmodel = (
+            BenchmarkModel.objects.filter(benchmark__id=benchmark.id, model_mlcube__id=mlcube.id)
+            .order_by("-created_at")
+            .first()
+        )
+        if not last_benchmarkmodel:
```
aristizabal95, line 23: Current implementation raises an error when the benchmark is executed with the reference model, since this model is not associated to the benchmark through the `BenchmarkModel` object.

aristizabal95: We would need an if in the lines of 
`if not last_benchmarkmodel and not referencemodel:`
I'm guessing this would require retrieving the current benchmark

aristizabal95 (changes requested):
Hey Johnu, just a minor bug coming up here. I'd like to know if you think my suggestion is okay, or if another fix should be implemented

aristizabal95:
I modified the CLI script so it approves the dataset association

server/result/serializers.py
johnugeorge: Do we upload results for reference model as well?

aristizabal95: Yes we do upload results for reference models

johnugeorge: Which dataset is used for this result upload?

cli/cli.sh
```diff
@@ -66,6 +66,14 @@ fi
 echo "\n"
 DSET_UID=$(medperf --storage=$MEDPERF_STORAGE --host=$SERVER_URL dataset ls | tail -n 1 | tr -s ' ' | cut -d ' ' -f 1)
 echo "====================================="
+echo "Approving dataset association"
+echo "====================================="
+# Log in as the benchmark owner
+BENCHMARK_OWNER_TOKEN=$(curl -s -X POST "http://127.0.0.1:8000/auth-token/" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{  \"username\": \"testbenchmarkowner\",  \"password\": \"test\"}" | jq -r '.token')
+# Mark dataset-benchmark association as approved
+MODEL_EXECUTOR1_IN_BENCHMARK_STATUS=$(curl -s -X PUT "http://127.0.0.1:8000/datasets/1/benchmarks/1/" -H  "accept: application/json" -H  "Authorization: Token $BENCHMARK_OWNER_TOKEN" -H  "Content-Type: application/json" -d "{  \"approval_status\": \"APPROVED\"}" | jq -r '.approval_status')
```
johnugeorge, line 74: Change the var name ?

aristizabal95: I wasn't passing the server url variable to those commands. Should work now

aristizabal95 (approved):

aristizabal95:
Things are looking good now!

PR #80, opened 2022-03-02
Cli benchmark subcommands

### LINKED ISSUE #42: Running benchmark test flow on the CLI
Testing is a means for verifying that the benchmark execution flow runs successfully, and is required for an association request to be created. Additionally, testing can be useful for the development process, in which an user is creating a model and wants to ensure it can be executed even before creating the MLCube instance on the server.

Some implications of this:
- We need access to a public dataset for executing the flow. See #41 .
- We need to allow users to build local temporary mlcubes for testing their implementation.
- Running this would probably require running the data preparation flow, while not registering the demo dataset. We need some way of keeping the prepared dataset as temporary, or at least indicating that the dataset is for demo purposes and should not be registered

### LINKED ISSUE #58: use local cubes, not registered on the platform. (Required for next task specifically)

### LINKED ISSUE #77: Implement and move benchmark commands to subcommand

aristizabal95:
This PR adds all commands related to benchmark entities. Specifically, it implements the following commands:
```
medperf benchmark ls
medperf benchmark submit
medperf benchmark associate
medperf benchmark test
medperf test
```
As noted above, it also adds the test command, which is required to auto-fill some of the fields for creating a benchmark.
Closes #77. Closes #42. Closes #58.

What's missing ATM:
Unit testing for the new features


hasan7n (commented):
Looks good!

I noticed that `entities/registration.py` and `entities/dataset.py` have a lot of logic in common: 

1. `Registration.upload` and `Registration.request_approval` methods are never used anywhere, `Dataset.upload` and `Dataset.request_registration_approval` are used instead.
2. `Registration.write` is basically `Dataset.set_registration`.
3. `Registration.todict` is basically `Dataset.registration` property.

I will be merging these two entities or at least do some refactoring.

Note: just noticed that my comments are not specific to this branch's commits. We can safely merge anyway.

hasan7n (approved):

PR #101, opened 2022-03-23
Cli Association Compatibility Tests

### LINKED ISSUE #90: Provide input to commands through CLI arguments
Currently all the commands that require user input use the `input()` command to retrieve such information. We should also support providing that information through CLI arguments so users can pass such information when they run the command. 

This also includes approvals, which would allow users to automate approvals for each command, allowing them just to write the initial command and let everything run on the background.

### LINKED ISSUE #104: Benchmark models should not be cached
Right now, a benchmark object is cached. This includes the list of models associated to the benchmark. Since this data can change without current user's interaction, it should get updated every time the user requests that benchmark.

### LINKED ISSUE #111: Bug: os symlink failing when running the compatibility test

aristizabal95:
This PR implements running the compatibility tests when a user wants to associate a resource to a benchmark, and adds the test results to the association request.

Depends on #80 
Closes #111.
Closes #104.
Closes #90


aristizabal95:
Commit [8a96213](https://github.com/mlcommons/medperf/pull/101/commits/8a96213c80e1eebac75893143ef02dc7b1dd0736) fixes symlink error. Closes #111

aristizabal95:
Commit [fa1e85e](https://github.com/mlcommons/medperf/pull/101/commits/fa1e85e61026999de28969055c835ad3fbde375b) updates local benchmark cache if not running the tests. Closes #104.

hasan7n (commented):
I created two PRs to be merged before merging this one:
This PR #216 for minor edits, and #215 to close issue #214.

Two additional points I think we should address:

1. In `entities/registration.py`, I think this error should be reported in logging, not for the user in the UI:

https://github.com/mlcommons/medperf/blob/775f674144048aed632a628b1bfbfa390f809fcc/cli/medperf/entities/registration.py#L172-L188

Also this error will never be thrown according to the current code, but it would be good to keep it for future unintended mistakes. I suggest to turn it into logging and print for the user another message that says for e.g. "medperf has encountered an internal error...". The current message gives the impression that the user did something wrong.





2. The registration approval status for the dataset can take the value "REJECTED". For now, it seems that a rejected status will never be written into a registration file locally. The status gets updated in runtime but not in the storage. But if this changes in the future, line 169 below might cause a serious problem:
https://github.com/mlcommons/medperf/blob/775f674144048aed632a628b1bfbfa390f809fcc/cli/medperf/entities/dataset.py#L163-L184

I suggest for now to just edit line 169. But we should discuss about whether it makes sense to update the registration file with "rejected" status when the user rejects. Currently it stays "pending"

PR #108, opened 2022-04-06
Store predictions in unique folder per run

### LINKED ISSUE #107: MLCube predictions collide for same model execution
Since recent changes, were mlcubes Inputs/outputs are folders, there's a high chance that running the same model on different datasets will cause predictions to combine for each run. This is extremely problematic, as metrics will be calculated for the whole predictions, and will most probably corrupt the results.

Previously, there was a discussion for keeping every execution input/output in a separate folder. This would be the best solution for this issue

Thanks @hasan7n for reporting this.

aristizabal95:
This PR includes a small fix for issue #107 by storing predictions in separate folders for each dataset-model pair. Closes #107


msheller (approved):
I see no issues. Thanks, Alejandro!

PR #118, opened 2022-04-13
Adding Readme on SQL folder 

johnugeorge:
Added Readme to SQL folder


aristizabal95 (approved):
Looks good! Fixed the screw-up from before. Checks should go through now

PR #125, opened 2022-04-18
Update vulnerable dependencies

aristizabal95:
This PR combines all suggestions from dependabot to fix dsicovered vulnerabilities in PyYAML and Django


johnugeorge:
this has version dependency with mlcube?

msheller (approved):
Thanks, Alejandro!

aristizabal95:
> this has version dependency with mlcube?

We could remove it and instead ask users to install it manually beforehand

johnugeorge:
> > this has version dependency with mlcube?
> 
> We could remove it and instead ask users to install it manually beforehand

But, users would get conflict when try to install ?  We should ask MLCube team to update their package

aristizabal95:
> > > this has version dependency with mlcube?
> > 
> > 
> > We could remove it and instead ask users to install it manually beforehand
> 
> But, users would get conflict when try to install ? We should ask MLCube team to update their package

Agreed, I just asked Jurado. I guess we'll have to wait until they release a new update

aristizabal95:
That was faster than expected!

PR #126, opened 2022-04-18
Fix CI Linter silently failing

### LINKED ISSUE #75: CI Linter shouldn't silently fail
CI Workflow checks if code follows flake8 linter standards, but will silently fail if code isn't formatted correctly. This step should cause the workflow to fail, and stop PRs from being merged.

aristizabal95:
currently, flake8 silently fails on each PR. This fixes it by causing execution error if linter finds warnings or errors. Closes #75 

The important changes are inside `.github/local-ci.yml`, everything else is formatting changes to follow PEP8 standards


aristizabal95:
@johnugeorge Linter is failing due to complexity on some of the server validations. Would you be able to simplify them for this commit? Or should we temporarily increase the allowed complexity?

johnugeorge (approved):

PR #127, opened 2022-04-20
Add path validation to dataset create

### LINKED ISSUE #64: Passing garbage paths to `medperf prepare` should fail
When executing `medperf prepare`, passing nonsensical paths to the `-d` and `-l` paths should raise an error, but instead it runs without any problem.

example:
```
medperf prepare -b 1 -d 1 -l 1
```

Some things to note:
- When executing the flow, folder `1` (for the example above) is created on the cwd. It's just an empty directory
- The resulting prepared dataset contains no data, but is able to generate statistics (from where?)

aristizabal95:
This small PR fixes a bug where passing garbage paths to `medperf dataset create` would create new paths in the current directory and run the workflow without issues. Closes #64


hasan7n (approved):
Looks great!

PR #128, opened 2022-04-21
Cli offline testing

### LINKED ISSUE #110: Support for "offline" or "unauthenticated" modes
For FeTS challenge (and future challenges), we want to enable competitors to test the challenge pipeline without having MedPerf user accounts. Initial thoughts on how this could work:

1. Use Mocker and have specific mocked servers for given challenges
2. Actual local server with pre-configured db
3. Mechanism to specify a pipeline directly (i.e. no benchmark required) with pre-cached cubes

aristizabal95:
This PR provides a mechanism for allowing offline testing if all required assets for a test are already present at the user's computer. 
It does this by requiring authentication only if an authenticated request is ever executed. This is different from before, where we checked authentication at the beginning of the commands that required it, which blocked users from using any of the commands without a valid `credentials` file inside the medperf storage. 
Closes #110.
Depends on #101


msheller (approved):
Looks good! Thanks, Alejandro!

PR #131, opened 2022-04-25
Recreate remote dataset on duplicate submission

### LINKED ISSUE #132: Update local dataset to reflect remote on submission.

aristizabal95:
This PR checks for current-user remote datasets with the same registered uid as the dataset being submitted. If there's a match, then the local dataset is updated so it contains information of the remote dataset. Fixes #132


hasan7n:
We should now expect users to enter the full dataset uid when using `medperf dataset submit -d <uid>`.
If a uid hint was entered, the local dataset will not be updated, and the "submission exists" error message will be printed.
If we disable uid hinting, the error printed will make more sense: local dataset is not found.
In this case we should make sure uid hints are not accepted anywhere (e.g. in compatibility test)

hasan7n:
a much better alternative to disabling uid hint is to simply apply the logic introduced in this PR using the full data uid after being constructed by the `Dataset` class.

hasan7n (approved):

PR #142, opened 2022-05-10
CLI MLCube Timeout

### LINKED ISSUE #140: [FEATURE] MLCube timeout option
We'd like to add a simple time-out option for MLCubes to the medperf commands. Ideally, we could specify different timeouts for different pipeline stages. All this needs to do is kill the container if it hasn't returned in time and then return the timeout error.

This will let us ensure competitor models aren't going to take so long that the hospitals cannot complete their evaluations in time.

aristizabal95:
This PR adds functionality for specifying timeouts for each task independently. Closes #140.


aristizabal95:
@johnugeorge it seems we need to configure something to execute the integration workflow execution from a forked PR

Issue #95, opened 2022-03-15
Benchmark ls is failing when getting user benchmarks


aristizabal95:

aristizabal95:
At the time of creating this issue I didn't have much information about what conditions caused this error. I wasn't able to reproduce locally

hasan7n:
The error was:
```
MedPerf 0.0.0
Traceback (most recent call last):
  File "/home/hasan/anaconda3/envs/medperf3.9/bin/medperf", line 33, in <module>
    sys.exit(load_entry_point('medperf', 'console_scripts', 'medperf')())
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/typer/main.py", line 214, in __call__
    return get_command(self)(*args, **kwargs)
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/hasan/anaconda3/envs/medperf3.9/lib/python3.9/site-packages/typer/main.py", line 500, in wrapper
    return callback(**use_params)  # type: ignore
  File "/home/hasan/Desktop/tasks/medperf/cli/medperf/commands/benchmark/benchmark.py", line 24, in list
    BenchmarksList.run(comms, ui, all)
  File "/home/hasan/Desktop/tasks/medperf/cli/medperf/commands/benchmark/list.py", line 26, in run
    for bmark in benchmarks:
TypeError: 'NoneType' object is not iterable
```


[...]

PR #146, opened 2022-05-15
fix upload benchmark

hasan7n:
`medperf/cli/medperf/comms/rest.py: REST.upload_benchmark` should return benchmark id. As reported in #95


aristizabal95 (commented):
Simple yet effective. Thanks Hasan for this fix! I'll merge it right away

PR #159, opened 2022-05-31
Add support for different MLCube platforms

aristizabal95:
This PR adds support to different MLCube platforms through the `--platform` parameter. This parameter is then passed to the respective mlcubes.


msheller (approved):
Looks good. Glad to see it was such a simple change!

PR #185, opened 2022-06-16
[WIP] Enforce https. Add certificate argument

aristizabal95:
This PR adds functionality for enforcing https (by hardcoding the protocol inside the REST instance) and provides an argument for indicating the path to a certificate if necessary.

Work In Progress. This commit breaks connection to locally hosted servers, as they run in http by default. Changes required on the server

### DIFF
--- a/cli/medperf/comms/rest.py
+++ b/cli/medperf/comms/rest.py
@@ -37,9 +37,7 @@ class REST(Comms):
             pwd (str): Password
         """
         body = {"username": user, "password": pwd}
-        res = requests.post(
-            f"{self.server_url}/auth-token/", verify=self.cert, json=body
-        )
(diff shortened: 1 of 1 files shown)


cli/medperf/comms/rest.py
```diff
@@ -25,7 +37,9 @@ def login(self, ui: UI, user: str, pwd: str):
             pwd (str): Password
         """
         body = {"username": user, "password": pwd}
-        res = requests.post(f"{self.server_url}/auth-token/", json=body)
+        res = requests.post(
```
aristizabal95, line 40: We also need to catch errors here

aristizabal95:
Reference for generating certificates: https://github.com/intel/openfl/blob/develop/openfl/cryptography/ca.py

PR #189, opened 2022-06-23
Ensure SSL support for local deployment

johnugeorge:
This PR adds

1. SSL certificates are mandatory for local deployment. We cannot use `python manage.py runserver` anymore
2. For local deployment/testing,  just run `sh setup-dev-server.sh -c cert.crt -k cert.key`  which automatically creates a self signed certificate `cert.crt` with key `cert.key` and starts the SSL server at https://127.0.0.1:8000 
  To use seed.py for self signed server certificates, run `python seed.py --cert <path to generated cert.crt file>`


server/mlcube/tests.py
```diff
@@ -10,6 +11,11 @@ class MlCubeTest(TestCase):
     """Test module for MLCube APIs"""
 
     def setUp(self):
+        # Disable SSL redirect in tests
```
aristizabal95, line 14: it'd be nice if there's a way to not repeat this configuration.

server/setup-dev-server.sh
```diff
@@ -0,0 +1,56 @@
...
+  echo "KEY FILE must not be empty"
+  exit 1
+fi
+
+if [ "$CERT_GENERATE" -eq 0 ]
+then
+  echo "No certs are generated as CERT_GENERATE flag is disabled"
+else
+    echo "Certs are generated"
+    openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout cert.key -out cert.crt -subj "/C=US/ST=Any/L=Any/O=MedPerf/CN=127.0.0.1" -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
```
aristizabal95, line 38: 3072 is the recommended RSA bit-size by Micah

server/user/tests.py
```diff
@@ -7,6 +8,11 @@ class UserTest(TestCase):
     """Test module for users APIs"""
 
     def setUp(self):
+        # Disable SSL redirect in tests
+        settings_manager = override_settings(SECURE_SSL_REDIRECT=False)
```
aristizabal95, line 12: Same as previous comments. Just to ensure this setting is. consistent everywhere even if we need to make changes in the future

aristizabal95 (commented):
Looks good! I really like the script you provided to configure the dev server. It's going to be really useful, thanks! I'll see if I can bring my changes to this branch.

server/user/tests.py
johnugeorge: Moved to common class

aristizabal95:
@johnugeorge I implemented my changes

server/setup-dev-server.sh
```diff
@@ -0,0 +1,59 @@
...
+  echo "KEY FILE must not be empty"
+  exit 1
+fi
+
+if [ "$CERT_GENERATE" -eq 0 ]
+then
+  echo "No certs are generated as CERT_GENERATE flag is disabled"
+else
+    echo "Certs are generated"
+    openssl req -x509 -nodes -days 365 -newkey rsa:3072 -keyout cert.key -out cert.crt -subj "/C=US/ST=Any/L=Any/O=MedPerf/CN=127.0.0.1" -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
```
johnugeorge, line 40: Why do we need both crt and key file? We just need .crt file for client

`curl https://127.0.0.1:8000/ --cacert cert.crt `
`requests.request(url="https://127.0.0.1:8000/", method='get', verify='cert.crt')`

both work

johnugeorge: Is the validation failing for you even providing crt file?

aristizabal95: I wasn't aware requests could use the .crt file for verification. That was ignorance from my side, I'll make the required changes to pass the .crt instead of the .pem file I added

johnugeorge: @aristizabal95 , One question: Why can't we reuse the variable CERT_FILE var? Do we need to add extra parameter `CERT_PATH` as it is the same cert file?  In Line 41, cert.key and cert.crt needs to be changed to $CERT_FILE and $KEY_FILE respectively as well

With that, isn't this good enough to generate certs? `sh setup-dev-server.sh -c ~/.medperf_test.crt -k ~/.medperf_test.key`

aristizabal95: Agreed. Then we need to use the $CERT_FILE and $KEY_FILE inside the openssl command
```suggestion
    openssl req -x509 -nodes -days 365 -newkey rsa:3072 -keyout $KEY_FILE -out $CERT_FILE -subj "/C=US/ST=Any/L=Any/O=MedPerf/CN=127.0.0.1" -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
```
And we can remove the additional parameter and extra copy

msheller (approved):
Looks great, Johnu! Thanks so much for doing this so quickly.

cli/cli.sh
```diff
@@ -47,38 +51,38 @@ ls $DIRECTORY/mock_chexpert/valid
...
-  cat "$MEDPERF_STORAGE/medperf.log"
+  tail "$MEDPERF_LOG_STORAGE"
   exit 1
 fi
 echo "\n"
 echo "====================================="
 echo "Running data preparation step"
 echo "====================================="
-echo ${LOCAL:+'-e'} "Y\nname\ndescription\nlocation\nY\nY\n" | medperf --host=$SERVER_URL --log=DEBUG --storage=$MEDPERF_STORAGE dataset create -b 1 -d $DIRECTORY/mock_chexpert -l $DIRECTORY/mock_chexpert
+echo ${LOCAL:+'-e'} "Y\nname\ndescription\nlotailion\nY\nY\n" | medperf --certificate $CERT_FILE --host=$SERVER_URL --log=DEBUG --storage=$MEDPERF_STORAGE dataset create -b 1 -d $DIRECTORY/mock_chexpert -l $DIRECTORY/mock_chexpert
```
johnugeorge, line 64: Typo

PR #190, opened 2022-06-24
Add DB schema changes 

johnugeorge:
1. Remove uniqueness constraint of tuple `("benchmark", "model", "dataset")` on results 
2. Add extra fields `image_tarball_url` and `image_tarball_hash`
3. Rename existing fields `tarball_url` and `tarball_hash` to `additional_files_tarball_url` and `additional_files_tarball_hash` respectively.


aristizabal95:
After discussing with the Best Practices team, we realized that the image might not always come as a tarball file, and even if it is a tarball, should not be treated as such by medperf (we shouldn't try and untar it). Singularity images will show up as .sif, while docker images will be tarballs that are expected as such by docker.
I bring this up to suggest:
- New fields should be named `image_url` and `image_hash`
The other fields could still be renamed for readability's sake, although we could also keep them for backwards-compatibility. Let me know what you think @johnugeorge and @msheller

johnugeorge:
So, this one?

tarball_url renamed to additional_files_tarball_url
tarball_hash renamed to additional_files_tarball_hash
image_url(new field)
image_hash(new field)

cli/medperf/commands/mlcube/submit.py
```diff
@@ -110,9 +128,12 @@ def todict(self):
             "state": "OPERATION",
             "is_valid": True,
         }
+        if self.image_file:
+            dict["image_tarball_url"] = self.image_file
+            dict["image_tarball_hash"] = self.image_tarball_hash
```
msheller, line 133: At some point, we should change "image_tarball" to just "image" throughout this class for clarify.

cli/medperf/tests/entities/test_cube.py
```diff
@@ -12,8 +12,10 @@
 PATCH_CUBE = "medperf.entities.cube.{}"
 CUBE_PATH = "cube_path"
 PARAMS_PATH = "params_path"
-TARBALL_PATH = "tarball_path"
-TARBALL_HASH = "tarball_hash"
+TARBALL_PATH = "additional_files_tarball_path"
+TARBALL_HASH = "additional_files_tarball_hash"
+IMG_PATH = "image_tarball_url"
+IMG_HASH = "image_tarball_hash"
```
msheller, line 18: Another spot we should eventually clean up to just read "image" instead of "image_tarball"

server/mlcube/migrations/0002_auto_20220624_0853.py
```diff
@@ -0,0 +1,37 @@
...
+        migrations.RenameField(
+            model_name='mlcube',
+            old_name='tarball_url',
+            new_name='additional_files_tarball_url',
+        ),
+        migrations.AddField(
+            model_name='mlcube',
+            name='image_tarball_hash',
+            field=models.CharField(blank=True, max_length=100),
+        ),
```
msheller, line 27: shouldn't these read "image_hash" and "image_url" instead of "image_tarball..."? I may be confused here.

msheller (commented):
I may be confused: I'm seeing "image_tarball_..." instead of "image_..." in many places.

Otherwise, looks good to me!

aristizabal95:
@msheller I've been talking with the Best Practices WG and I couldn't find sufficient reasons to argue in favor of having the image not compressed into a tarball given the possibility of future supported image files that might require multiple files in the same place. We can definitely change the name to image, although tarball is more descriptive to what we expect to see stored in those fields

msheller (approved):
I like Alejandro's explanation of just assuming image files will always be tarballs. Looks good, thanks guys!

PR #191, opened 2022-06-24
Adding Password change api

johnugeorge:
Added  new api  `POST /me/password/` to change own user's password. This API is open to all authenticated users. 

It takes one argument `password` and returns success message (200 code) if succeeded.

If password is changed successfully, user token will be deleted automatically. The user needs to login again to get a new token for any further api calls.


cli/medperf/commands/auth/password_change.py
```diff
@@ -0,0 +1,27 @@
...
+        pwd_repeat = ui.hidden_prompt("Please retype your new password: ")
+        if pwd != pwd_repeat:
+            pretty_error(
+                "The passwords you typed don't match. Please try again.",
+                ui,
+                clean=False,
+                add_instructions=False,
+            )
+
+        comms.change_password(pwd)
```
msheller, line 24: What if this fails?

cli/medperf/comms/rest.py
```diff
@@ -31,6 +31,18 @@ def login(self, ui: UI):
...
+
+        Args:
+            pwd (str): New password to be set
+            ui (UI): Instance of an implementation
+        """
+        body = {"password": pwd}
+        res = self.__auth_post(f"{self.server_url}/me/password/", json=body)
+        if res.status_code != 200:
+            pretty_error("Unable to change the current password", self.ui)
+
```
msheller, line 45: Right now, we're relying on a specific detail of this implementation to prevent deleting the credential in the case of this API failing. I think it would be best if this interface explicitly returned a boolean for us to check, even if in our current implementation it will never matter (because we error out here).

cli/medperf/tests/commands/auth/test_password_change.py
```diff
@@ -0,0 +1,61 @@
+import pytest
+from unittest.mock import call, ANY
+
+from medperf.comms import Comms
+from medperf.commands.auth import PasswordChange
+from medperf.utils import storage_path
+from medperf import config
+
```
msheller, line 8: Possible to add a failure test that checks that the token is NOT removed if change password fails?

msheller (changes requested):
Looks good! Small change request:

I think we need to ensure that the change password call in the CLI explicitly returns false on failure. Currently, it would be possible to use a different comms module that doesn't close the process via an error and end up deleting the token even though the password was not changed.

Should be a super simple change. A test for this case would nice helpful, if it's easy enough (testing that if the password change call fails then the token is not deleted).

cli/medperf/comms/rest.py
aristizabal95: Thanks Micah, this makes sense. Checking both sides ensures that future comms implementations that might not explicitly error out will be handled correctly either way

cli/medperf/commands/auth/password_change.py
aristizabal95: I just added the check for this situation

cli/medperf/tests/commands/auth/test_password_change.py
aristizabal95: Added, thanks!

msheller (approved):
Looks good! Thanks for the changes, Alejandro!

PR #204, opened 2022-07-14
Hash checking for cached mlcubes

### LINKED ISSUE #202: [BUG] Hashes are not checked if MLCube is already cached
**Describe the bug**
During the execution of an MLCube, we want to ensure the hash of the downloaded files matches the hashes provided at submission time. This is done when the MLCube is retrieved for the first time. But, if the MLCube has already been downloaded, this step is being skipped somehow.

**To Reproduce**
1. Submit a cube with additional files or image file. Create a benchmark that uses that cube
2. Modify the contents of the link used for the file so the hash changes
3. Try to run the benchmark. It should fail because of hash mismatch
4. Run the benchmark again. Execution should now continue without issues

**Expected behavior**
Execution of an MLCube with hash mismatch should never work

**Additional context**
This is being caused because of the following reasons:
1. **The mlcube metadata is not being stored**
  The communication interface returns the contents of the metdata (dict) and that is used at retrieval time to obtain all the necessary files and their corresponding hashes. This metadata dictionary is then disposed after execution
2. **When retrieving local mlcubes, the wrong mlcube file is being used for metadata**
  https://github.com/mlcommons/medperf/blob/22c0838e6bf2404a07344e5f7e17b63cd5de64d2/cli/medperf/entities/cube.py#L80-L82
  This is loading the `mlcube.yaml` file instead of any actual metadata like what is obtained from the server. This is a major issue that should be fixed ASAP.
3. **The local "metadata" dictionary no longer contains information about additional files**
  https://github.com/mlcommons/medperf/blob/22c0838e6bf2404a07344e5f7e17b63cd5de64d2/cli/medperf/entities/cube.py#L144-L146
  Since the wrong metadata contents are being used, the cube instance believes it doesn't have additional files or images, and therefore it doesn't check any hash

**Possible Fix**
The best fix I can think of is:
- At retrieval time, get all the information as is currently being done. For additional files and images, get their hash and store that hash in the `meta` dictionary
- Store the metadata file inside the mlcube directory. This can be inside an `mlcube-meta.yaml` file (name should be stored in the `config.py` file)
- When retrieving local mlcubes, load the correct metadata contents
- When checking mlcube validity, compare the hash that was originally provided from the metadata against the computed hash (which should also be stored in the `mlcube-meta.yaml` file)

hasan7n:
This PR allows to perform hash check of a cached MLCube each time it is retrieved from local storage. This will be done by comparing server hashes and locally computed hashes which are cached during the MLCube retrieval.

Closes #202

### DIFF
--- a/cli/medperf/commands/compatibility_test.py
+++ b/cli/medperf/commands/compatibility_test.py
@@ -174,6 +174,16 @@ class CompatibilityTestExecution:
             dst = os.path.join(cubes_storage, temp_uid)
             os.symlink(path, dst)
             logging.info(f"local cube will be linked to path: {dst}")
+            cube_metadata_file = os.path.join(path, config.cube_metadata_filename)
+            cube_hashes_filename = os.path.join(path, config.cube_hashes_filename)
+            if not os.path.exists(cube_metadata_file):
+                metadata = {"name": temp_uid, "is_valid": True}
+                with open(cube_metadata_file, "w") as f:
+                    yaml.dump(metadata, f)
+            if not os.path.exists(cube_hashes_filename):
+                hashes = {"additional_files_tarball_hash": "", "image_tarball_hash": ""}
+                with open(cube_hashes_filename, "w") as f:
+                    yaml.dump(hashes, f)
             return
 
         logging.warning(f"mlcube {val} was not found as an existing mlcube")
--- a/cli/medperf/config.py
+++ b/cli/medperf/config.py
@@ -35,6 +35,7 @@ cube_submission_id = "tmp_submission"
 test_dset_prefix = "test_"
 demo_dset_paths_file = "paths.yaml"
 cube_metadata_filename = "mlcube-meta.yaml"
+cube_hashes_filename = "mlcube-hashes.yaml"
 
 default_comms = "REST"
 default_ui = "CLI"
--- a/cli/medperf/config.py
+++ b/cli/medperf/config.py
@@ -34,6 +34,7 @@ test_cube_prefix = "test_"
 cube_submission_id = "tmp_submission"
 test_dset_prefix = "test_"
 demo_dset_paths_file = "paths.yaml"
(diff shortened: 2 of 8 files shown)


aristizabal95 (commented):
Hey Hasan, it's looking great! Can you add related tests for the new function and to ensure invalid local are detected by the `is_valid` function? After that I think this PR is ready

aristizabal95:
Also, we would like to encourage PRs to have some details of what has been done. Can you also add a brief description? Thanks!

hasan7n:
This PR will cause compatibility tests to fail when given a local path to an mlcube if the mlcube doesn't have the new two files (metadata file and hashes file).

Cached mlcubes are now expected to have `mlcube-meta.yaml` and `mlcube-hashes.yaml` files when retrieved by `Cube.all`.
Hash checking for mlcubes that are given locally by users should be disabled.

**Possible fix**

The compatibility test command will create a symbolic link for the local mlcube paths as follows:
https://github.com/mlcommons/medperf/blob/4b02ab8e395ace19bad4bf5d038df61a9f5860ee/cli/medperf/commands/compatibility_test.py#L169-L177
Locally given cubes have unique UID patterns.
A fix can be to disable hash checking when the cube UID starts with `config.test_cube_prefix`. In this case the new files don't need to exist, and a minimal `meta` dictionary can be created with empty hashes when retrieving the cube in `Cube.all`.

Another way can be to create an `mlcube-meta.yaml` and `mlcube-hashes.yaml` with empty hashes when the symbolic link is created. This can be done:

- By directly adding the file to the local cube, or
- By copying the local cube to the medperf storage instead of using a symbolic link, and adding the file there.


**A possible human-error**
Users might give the compatibility test command a cached mlcube that was retrieved from the server.

**possible fix**
Check first if the locally given mlcube has the metadata and hash files before proceeding with no hash checking.

aristizabal95 (approved):
Great work Hasan, thanks for this!

PR #205, opened 2022-07-14
Invalid MLCube Check

### LINKED ISSUE #29: Invalid MLCubes should not be executed by the CLI
Cubes can be invalidated due to security vulnerabilities. If that's the case, we should default to not running the cube.

aristizabal95:
This PR adds a check for the `is_valid` field before execution. This ensures that MLCubes that have been marked invalid won't be executed by the platform. Closes #29.


hasan7n (approved):
Looks good!

PR #211, opened 2022-07-19
Pass metadata through the generated_metadata key

### LINKED ISSUE #210: [BUG] Dataset Metadata is not being passed correctly to the server
**Describe the bug**
A while ago, the `metadata` field for datasets was changed to `generated_metadata`. This change was not reflected on the CLI, which is causing new datasets to be stored without the metadata or statistics generated by the Data Preparator MLCube

**To Reproduce**
Prepare a dataset and look at the contents of the dataset object in the server

**Expected behavior**
The statistics should be passed through the `generated_metadata` key.

aristizabal95:
This PR fixes the issue where the dataset statistics were not being stored in the server because of key mismatch. Closes #210

### DIFF
--- a/cli/medperf/entities/dataset.py
+++ b/cli/medperf/entities/dataset.py
@@ -50,7 +50,11 @@ class Dataset:
         self.input_data_hash = registration["input_data_hash"]
         self.separate_labels = registration.get("separate_labels", False)
         self.split_seed = registration["split_seed"]
-        self.generated_metadata = registration["metadata"]
+        if "metadata" in registration:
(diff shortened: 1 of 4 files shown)


hasan7n (commented):
Looks good!

Just a note: the metadata key in the `registration.py` file is still "metadata". This will not cause errors since the change is backward compatible, but the key will always be "metadata" the first time the dataset is registered. I suggest to change that too.

aristizabal95:
Hey @hasan7n I just added your request. Thanks for that! please check the recent changes

hasan7n (approved):

PR #215, opened 2022-07-28
Fix benchmark associate with no entities

### LINKED ISSUE #214: [BUG] Benchmark associate command works without providing neither a cube nor a dataset
**Describe the bug**
Benchmark association with no cube or dataset will show the user that it is successfully done. This can be misleading in case the user forgot to provide a cube or a dataset but thought that they did. The code actually will do nothing.

**To Reproduce**
Run `medperf benchmark associate -b 1`
You will see a "Done!" message.

**Expected behavior**
An error should be thrown in case no dataset and no cube were provided to the command.

hasan7n:
This PR prevent `medperf benchmark associate` command to be run without providing neither a model nor a dataset.
Closes #214

### DIFF
--- a/cli/medperf/commands/benchmark/associate.py
+++ b/cli/medperf/commands/benchmark/associate.py
@@ -26,10 +26,11 @@ class AssociateBenchmark:
             ui (UI): Instance of UI interface
             approved (bool): Skip approval step. Defaults to False
         """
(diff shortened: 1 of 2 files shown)


aristizabal95 (approved):
Looks great! Thanks for this!

PR #221, opened 2022-08-09
Fix reject association throwing invalid value error

aristizabal95:
This PR fixes association rejection failing due to using `REJECT` instead of `REJECTED`.


hasan7n (approved):
Looks good!

PR #223, opened 2022-08-09
use enums for approval status

hasan7n:
The code now will work with python Enums for approval status (APPROVED, REJECTED, PENDING) instead of strings.


aristizabal95 (approved):
Looks great! Thanks for this change! This will ensure we are less error-prone in the future when it comes to handling approval status

PR #228, opened 2022-08-17
Reject benchmark execution and associations if the dataset has no server uid

### LINKED ISSUE #133: Reject benchmark execution if dataset has no uid

hasan7n:
This PR makes sure that benchmark execution (`medperf result create`) and associating datasets to benchmarks (`medperf dataset associate`) will fail if the dataset is not yet submitted to the server. Closes #133


aristizabal95 (approved):
Great! Quick and simple, yet effective. Thanks Hasan!

PR #234, opened 2022-08-24
MLCube image downloads at get time

### LINKED ISSUE #192: [FEATURE] Separate mlcube image download from mlcube execution
Currently, files for running an mlcube are downloaded before execution, but the mlcube image is downloaded by MLCube if not found beforehand. This works fine, except when you want to add a timeout to the mlcube execution, since the download time will be considered as execution time.

Downloading the mlcube image before execution would be ideal. This should be possible, as mlcube provides the `configure` command which allows downloading the image without executing it

aristizabal95:
This PR retrieves the mlcube image when executing `Cube.get()` if the image is stored in some container hub. Previously, the image would be downloaded at run-time, which could cause issues when using execution `timeout`, since the execution time would have to also consider download time. Closes #192


hasan7n (commented):
I am a bit doubtful about removing `untar(image_path)` here.
https://github.com/mlcommons/medperf/blob/c55da305f9054fbdda17dfe3bd4432071f567730/cli/medperf/entities/cube.py#L151-L156

Haven't tested though. Otherwise, everything looks good!

aristizabal95:
@hasan7n good catch! I must have thought of placing it somewhere else and forgot to add it again. I fixed the bug and I als ocreated a test to ensure it doesn't go undetected next time. Please give it another look

hasan7n (approved):
Thanks!

PR #235, opened 2022-08-26
Update default server url

### LINKED ISSUE #231: Host api at api subdomain api.medperf.org
We are finishing the website, which going to be hosted on the main domain webperf.org
and so need to move api for CLI to the api subdomain -> api.medperf.org

hasan7n:
This PR updates the default server url. Closes #231

### DIFF
--- a/cli/medperf/config.py
+++ b/cli/medperf/config.py
@@ -1,7 +1,7 @@
 from os.path import expanduser, abspath
 
 version = "0.0.0"
-server = "https://medperf.org"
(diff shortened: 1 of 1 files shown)


johnugeorge:
One thought. Is it better to make server url mandatory for the CLI than keeping a default ? if configured wrongly, central api server can be overloaded with requests.

aristizabal95 (approved):
Not much could go wrong here XD thanks for the quick PR!

PR #252, opened 2022-09-07
Fix Github pages custom domain being ignored

### LINKED ISSUE #250: [BUG] Custom domain for medperf documentation gets overwritten
The custom domain for medperf documentation (docs.medperf.org) is automatically changing after merging each pull request to main.
The problem must be because of `--force` when running `mkdocs`:

https://github.com/mlcommons/medperf/blob/71cba048727eb25172245f501e62b6017740580d/.github/workflows/docs-ci.yml#L16-L17

Setting a custom domain adds a file `CNAME` to `gh-pages` branch.
`--force` in `mkdocs` force-pushes the documentation changes to `gh-pages` branch, ignoring any existing commit.

It would be better to first test removing `--force` in a fork to make sure it will work properly. Until this is fixed, **we should manually set the custom domain** after merging a pull request, to avoid having an invalid link on our live website https://medperf.org.

hasan7n:
Since we are deploying github pages through github actions workflow using `mkdocs`, we need to make sure a `CNAME` file is present in the `docs` folder of the main branch, so that `mkdocs` copies this file to the `gh-pages` branch each time it updates the documentation website content. This `CNAME` file contains the custom domain and should be present in order to use the custom domain for github pages.

Closes #250

### DIFF
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+docs.medperf.org


aristizabal95 (approved):
Quick and easy fix. Thanks for looking into this!

PR #254, opened 2022-09-09
refactor registration entity

### LINKED ISSUE #245: merging Registration and Dataset entities
Currently `entities/Registration` and `entities/Dataset` have a lot of common logic. It is good to merge them to mitigate ambiguities for new contributors. Also, would be good to edit variable names of the different dataset UID types (input uid, generated uid, server uid) to clear ambiguities.

hasan7n:
This PR moves around the code of the `Registration` entity to better organize its logic.
The plan first was to merge its code with the `Dataset` entity. However, it seems that it would be better to merge with the`DataPreparation` command, since `Registration` deals with unprepared datasets, while `Dataset` represents a prepared dataset. This actually explains why there was a separate entity `Registration` at the first place.

Below is the summary of to where the `Registration` class methods have been moved:

1. Methods moved to `DataPreparation`, replacing the `DataPreparation.create_registration` method:
    *  `generate_uids`
    * `to_permanent_path`
    * `todict`
    * `write`

2. `__get_stats` is moved to `utils`
3. `request_approval` and `upload` are already present in `Dataset`
4. `is_registered` is removed, as it is not used anywhere. It was used before we allow recreating already prepared datasets (#145)

Closes #245


aristizabal95 (commented):
Hey Hasan,

This looks great! Refactoring the registration entity is something that has been on my radar for a while, and I love to see it being tackled. Thanks for this!

I have one doubt and possibly a change request before approving this. If I'm correct, you're getting the statistics by looking at the mlcube's default output path. I'd suggest finding a unique location to store the statistics for the data/mlcube tuple, even if temporary. 

The thing with using the default path is that this will be shared across executions (they'll always be inside the mlcube workspace), which could cause collisions. For example, if a prepared a previous dataset with one mlcube, and then tried to prepare another one but silently failed, then when retrieving the statistics for the second dataset we will accidentally use the statistics from the previous dataset.

Let me know if this makes sense. This was a failure I previously encountered with predictions, and is the reason why they're now stored on unique folders per execution.

hasan7n:
@aristizabal95 Totally agree! I will look into that.

hasan7n:
The statistics output path is now in the same directory with the `registration-info.yaml`.

I am a bit hesitant about keeping the file, as now there is redundant information (statistics are also in the registration info). We can delete it. What do you think @aristizabal95 ?

aristizabal95:
I'd say either delete the file, or keep the statistics separate from the registration information, following how we have two separate files for mlcubes as well. Either option is fine, and I agree we don't want to have duplicate information

aristizabal95 (approved):
Looks great! Thanks Hasan!

PR #257, opened 2022-09-20
Adding Google analytics to the docs website

bekinin:
Adding the google analytics to the doc.medperf.org so that we can track visitors in addition to the medperf.org

Plus simple feedback widget


aristizabal95 (approved):
Looks good! Thanks Dima for this!

PR #258, opened 2022-09-21
Remove randomness from CLI unit tests

aristizabal95:
Random tests were added with the purpose of having a wider range of testing. Completely random testing is not recommended, as it makes it harder to reproduce and catch bugs. Instead, Property-based testing is recommended for the same kind of purpose. Additionally, random test parameters are usually not supported with higher-level testing tools , like the ones provided by VSCode.

For this reason, this PR removes the use of `rand_l` as testing parametrization, and instead uses hardcoded examples for each parameter. With this mode of testing, any encountered special value that causes bugs should be added as an example.

In the future, frameworks like [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) should be considered for fully property-based testing.


cli/medperf/tests/commands/dataset/test_create.py
```diff
@@ -94,10 +93,8 @@ def exists(path):
         else:
             spy.asset_not_called()
 
-    @pytest.mark.parametrize("cube_uid", rand_l(1, 5000, 5))
-    def test_get_prep_cube_gets_prep_cube_if_provided(
-        self, mocker, preparation, cube_uid, comms, ui
-    ):
+    @pytest.mark.parametrize("cube_uid", [1776, 4342, 573])
+    def test_get_prep_cube_gets_benchmark_cube(self, mocker, preparation, cube_uid, comms, ui):
```
hasan7n, line 97: The test name here seems to be changed by mistake, right?

cli/medperf/tests/test_utils.py
```diff
@@ -153,7 +153,7 @@ def test_cleanup_removes_only_invalid_datasets(mocker, datasets):
 
 
 @pytest.mark.parametrize("path", ["path/to/uids", "~/.medperf/cubes/"])
-@pytest.mark.parametrize("datasets", rand_l(1, 1000, 2), indirect=True)
+@pytest.mark.parametrize("datasets", [4, 287], indirect=True)
```
hasan7n, line 156: There is still randomness when executing the `datasets` fixture defined on line 33. These numbers here are used to randomly sample UIDs in the pytest fixture.

cli/medperf/tests/test_utils.py
```diff
@@ -340,7 +340,7 @@ def test_approval_prompt_returns_approved_boolean(mocker, ui, input_val):
     assert approved == (input_val in "yY")
 
 
-@pytest.mark.parametrize("dict_with_nones", rand_l(0, 100, 5), indirect=True)
+@pytest.mark.parametrize("dict_with_nones", [32, 7, 90], indirect=True)
```
hasan7n, line 343: Same argument as the `datasets` fixture. `dict_with_nones` also contains randomness.

hasan7n (changes requested):
Thanks Alejandro!
I left a few comments let me know what do you think.
Also, we can remove `rand_l` definition from utils as it will not be used anymore.

cli/medperf/tests/commands/dataset/test_create.py
aristizabal95: You're right! I missed that one. This was most probably due to merge conflicts and didn't realize the other name was the correct one

cli/medperf/tests/test_utils.py
aristizabal95: Good point! I didn't look at the fixtures. I've removed those calls as well

aristizabal95:
Hey Hasan, thanks for the review! I applied your suggestions. Let me know if there's anything else I should modify

cli/medperf/tests/test_utils.py
```diff
@@ -32,7 +32,7 @@ def isdir(dir):
 @pytest.fixture
 def datasets(request):
     size = request.param
-    uids = rand_l(1, 5000, size)
+    uids = list(range(size))
     uids = [str(x) for x in uids]
     for i in range(size):
         if random.randint(0, 1):
```
hasan7n, line 38: line 38

hasan7n: and line 101

hasan7n (approved):
Thanks Alejandro for the changes!

I left a comment. There are two lines left having random behavior, sorry I think I missed them during the first look. But I think they are okay as they are traceable; not very random. Feel free to change them if necessary.

aristizabal95:
I totally missed those as well! I already addressed your comments, and I believe we were able to remove randomness from our testing!

PR #259, opened 2022-09-22
Implement Entities Interface

aristizabal95:
Most entities share some common functionality. Encapsulating this with an interface can make the code highly modular, and offers the possibility of adding commands to all entities with a single implementation. This PR provides the foundation to refactoring shared commands (`ls`, `submit`) and implementing new common commands (`edit`, `deploy`, `delete`) efficiently.


hasan7n:
Hey Alejandro! It looks nice, thanks for this. I am still reviewing it.

I saw that in some methods (.`get`, .`all`), `ui` and `comms` are taken directly from the `config`. Actually it looks better to do this. Do you think we should also follow this approach everywhere? To have, at the end, no `ui` or `comms` passed as parameters. Always use globally defined `config.ui` and `config.comms`. This is just for code consistency, we can make it a separate PR.

cli/medperf/entities/dataset.py
```diff
@@ -35,13 +34,16 @@ def __init__(self, data_uid: int, ui: UI):
...
+        if registration is None:
+            data_uid = self.__full_uid(data_uid, config.ui)
+            self.data_uid = data_uid
+            self.dataset_path = os.path.join(
+                storage_path(config.data_storage), str(data_uid)
+            )
+            registration = self.get_registration()
+        else:
+            self.data_uid = registration["generated_uid"]
         self.data_path = os.path.join(self.dataset_path, "data")
```
hasan7n, line 46: `self.dataset_path` doesn't seem to be defined if `registration` is not `None`.

cli/medperf/entities/dataset.py
```diff
@@ -26,7 +25,7 @@ class Dataset:
     data preparation output.
     """
 
-    def __init__(self, data_uid: int, ui: UI):
+    def __init__(self, data_uid: int, registration=None):
```
hasan7n, line 28: what about expecting the `__init__` to always recieve a registration. `uid`-based retrievals would happen through `.get` or `.all`. Maybe follow this convention in all entities?

hasan7n: So, the constructor of a Dataset will either receive the output hash (data_uid), or the registration. `Entity.get` will receive server uids only.

Do you think it would be better to split the constructor? something like abstract methods `Dataset.from_hash` and `Dataset.from_reg`.

aristizabal95:
Hey Hasan, thanks for giving this a look! I like all your ideas. Regarding `comms` and `ui` I absolutely agree, and was thinking of implementing this change everywhere in a separate commit, to keep this one "short".

As for the `__init__` I like your idea. We can keep `__init__` entity-specific, but implement a `from_dict()` method which creates an object given its dictionary description (e.g. `Benchmark.from_dict(benchmark.to_dict())` should create a copy).

Might be best to make this a separate PR?

hasan7n:
@aristizabal95 Alright! I will start another PR for the `__init__`.

aristizabal95:
@hasan7n about the review, do you think it's okay to merge? I can see a few issues that could benefit from this as well

hasan7n:
sure sorry... I thought we want to add the `ui` and `comms` thing here.

hasan7n (approved):

PR #260, opened 2022-09-22
Add rich for better help prompts

aristizabal95:
This PR is a simple quality-of-life improvement that provides better `--help` outputs by leveraging on `rich`.


aristizabal95:
Close because of using the wrong parent branch

PR #267, opened 2022-10-04
Implement output folder deletion after cube failure

### LINKED ISSUE #262: [BUG] CLI Doesn't remove dangling predictions from a failed execution
**Describe the bug**
In some cases, the prediction procedure may fail mid-execution. Usually, when failure happens, the `cleanup` procedure is executed to remove dangling files. This is not done with the failed predictions folder. This can be harmful because future executions of the same workflow will use an already-existing and probably occupied folder, which could cause side effects. Currently, the only way to overcome this is to ask the user to identify this situation and manually remove the conflicting folder. For example, in the FeTS Challenge, some participants encountered errors with an already-existing folders in the predictions output folder. If that folder was removed after failure, then the side effect would be avoided.

**Expected behavior**
If failure happens, identify the predictions folder of that specific execution and remove it inside `cleanup`. Using `--no-cleanup` will skip this, which allows for debugging.

aristizabal95:
This PR adds logic to handle unexpected errors during mlcube execution. If an error occurs, then the created output folder will be deleted to ensure future executions won't collide. Closes #262.


cli/medperf/commands/dataset/create.py
```diff
@@ -150,22 +151,33 @@ def run_cube_tasks(self):
...
+            self.ui.text = "Generating statistics..."
+            self.cube.run(
+                self.ui,
+                task="statistics",
+                timeout=statistics_timeout,
+                **statistics_params,
+            )
+            self.ui.print("> Statistics complete")
+        except RuntimeError as e:
+            cleanup_paths = [out_datapath, out_labelspath]
```
hasan7n, line 178: also `out_statistics_path`.
Maybe just use `self.out_path`?

cli/medperf/commands/dataset/create.py
```diff
@@ -150,22 +151,33 @@ def run_cube_tasks(self):
...
+            cleanup(cleanup_paths)
+            raise e
```
hasan7n, line 180: Why do we raise the error here, instead of using `pretty_error`?

in `result.create`:

```
except RuntimeError as e:
    logging.error(f"MLCube Execution failed: {e}")
    cleanup([preds_path, out_path])
    pretty_error("Benchmark execution failed", self.ui)
```

cli/medperf/decorators.py
```diff
@@ -24,7 +24,6 @@ def wrapper(*args, **kwargs):
         except Exception as e:
             logging.error("An unexpected error occured. Terminating.")
             logging.error(e)
-            cleanup()
-            raise e
+            pretty_error("An unexpected error occured")
```
hasan7n, line 27: missing `ui` :)

cli/medperf/commands/dataset/create.py
aristizabal95: Agreed, that would be easier!

cli/medperf/commands/dataset/create.py
aristizabal95: Ups just noticed the inconsistency. I agree, let's use `pretty_error` in all scenarios

cli/medperf/decorators.py
aristizabal95: I always forget that argument! Thanks for realizing!

hasan7n (approved):
Looks great! Thanks, @aristizabal95 

I added a test for `result.create` to also ensure cleanup will happen if benchmark execution failed. We are good to merge if the final modification looks good to you.

aristizabal95 (commented):
Awesome Hasan, thanks a lot for the help here!

PR #270, opened 2022-10-10
Remove comms and ui args from most calls

aristizabal95:
This PR aims to simplify methods calls by removing the `comms` and `ui` parameters, given that the ones assigned to the `config` singleton will always be used


hasan7n (approved):

hasan7n:
@aristizabal95 thanks for taking care of this daunting issue! Looks good to me.
also I removed ui/comms from the remaining utils

PR #272, opened 2022-10-10
Analogous Entities Implementations

### LINKED ISSUE #218: [BUG] MLCube submission ignores passed image hash
**Describe the bug**
When submitting an mlcube, one can pass the image-hash in order to skip image retrieval and hash computation. This is not happening, as no matter what, the hash is being computed during submission.

**To Reproduce**
1. Submit an mlcube specifying the `--image-hash` argument

**Expected behavior**
If passing the image hash manually, the image should not be downloaded and submission should be done with the passed hash

### LINKED ISSUE #271: [BUG] Dataset.get currently breaks
**Describe the bug**
https://github.com/mlcommons/medperf/blob/c90e7e96dbf7485f7668593f2810f4d36fd447d2/cli/medperf/entities/dataset.py#L64

The `Dataset` constructor currently expects `registration` to include a `status` key.

https://github.com/mlcommons/medperf/blob/c90e7e96dbf7485f7668593f2810f4d36fd447d2/cli/medperf/entities/dataset.py#L132-L133
`Dataset.get` above retrieves a registration dict from the server, which doesn't include the `status` key.

**To Reproduce**
`Dataset.get` is a recently added method. It is not used in default `medperf` commands yet.

hasan7n:
This PR addresses two issues regarding `medperf.entities`. 
1. This PR will bring the implementations of entities (`Dataset`, `Result`, `Benchmark`, and `Cube`) to become more analogous. 
2. This PR will ensure consistent server-client used data schema.

Both issues turns out to be dependent on each other and hence are both addressed here. This currently involves the following planned changes:

- Constructors of entities will be based on the dictionaries corresponding to their saved data in the server.
- Entities that can exist in the client before being registered in the database (`Dataset` and `Result`) will have an extra constructor as a class method based on their local identifier.
- `Entity.upload` will be modified to update the local dictionary instantly to reflect newly generated server data (e.g. `id`, `created_at`)
- `.write` will be implemented in all entities and added to the abstract `Entity` interface.
- `Entity.todict` will be designed to match the server data corresponding to the Entity. (Closes #271 )
- Figure out a way to use the same `Entitiy.todict` in both the entity definition (in `medperf.entities`) and the corresponding create/submit command (in `medperf.commands`) to prevent any future unintended errors caused by server-client data schema misalignment.

**UPDATE**:
Closes #218


cli/medperf/entities/benchmark.py
```diff
@@ -175,27 +187,30 @@ def todict(self) -> dict:
         dict: Dictionary containing benchmark information
         """
         return {
-            "uid": self.uid,
+            "id": self.uid,
```
aristizabal95, line 190: I think one of the next important steps is to somehow share this dictionary description of each entity so that we only have to specify it once. Right now we have the same structure copied multiple times across entities and commands. This is very error-prone. This is not specific to this PR, but something I realize we should work on soon given how this dictionaries increased in size.

cli/medperf/entities/cube.py
```diff
@@ -122,81 +121,77 @@ def get(cls, cube_uid: str) -> "Cube":
...
+        if self.git_parameters_url:
+            url = self.git_parameters_url
+            self.params_path = comms.get_cube_params(url, cube_uid)
+        if self.additional_files_tarball_url:
+            url = self.additional_files_tarball_url
             additional_path = comms.get_cube_additional(url, cube_uid)
-            additional_hash = get_file_sha1(additional_path)
+            if not self.additional_hash:
+                # log interactive ui only during submission
+                ui.text = "Generating additional file hash"
```
aristizabal95, line 146: I'm not sure how I feel about an entity directly interacting with the UI. If someone in the future wants to use medperf entities as an API, they'll encounter ui prints that could disrupt their workflows. I just realized this, and I'm aware we've been probably doing the same for a while with `pretty_error`s called by entities. 

It would be great to look into separating concerns here in the future.
- Entities are in charge of doing the hard work, and are unaware of an user interacting with them
- Commands are in charge of giving the user ways of interacting with entities, and therefore should provide user feedback.

cli/medperf/utils.py
```diff
@@ -487,3 +474,21 @@ def get_stats(data_path: str, remove: bool = True) -> dict:
...
+        path (str): path to the result root
+
+    Returns:
+        dict: results of the execution.
+    """
+    path = os.path.join(result_path, config.results_filename)
+    with open(path, "r") as f:
+        results = yaml.safe_load(f)
+
+    if remove:
```
aristizabal95, line 492: What are the use-cases of `remove=True`?

I believe it'd be best to separate this into two functions. `get_results` removing results doesn't seem intuitive

aristizabal95 (changes requested):
Hey Hasan! Amazing work! Things start looking cleaner and cleaner.
This Entity interface is going to be highlighting a lot of ways in which we can improve our code. I left a few notes regarding future enhancements, that don't need to be covered here.

I do want to know your opinion on separating `get_results` and `remove_results` before approving the PR, to keep separation of concerns there. Let me know if you think that's okay, or if there's something I'm missing about that implementation.

I didn't look into the tests, I trust they look good ;)

cli/medperf/entities/benchmark.py
hasan7n: Exactly.

I tried to avoid unintended future errors by designing the commands to directly call the entities whenever the command's `todict` is used (specifically inside `command.upload` and `command.write`). So currently entities constructors are making sure to have unified dictionaries between commands and entities. But yes, it's not the perfect way. I have added in this PR the expected dictionaries in `tests.mocks.requests`, maybe we can start from there...

cli/medperf/entities/cube.py
hasan7n: yeah.. maybe splitting up `cube.download` in the future would be a solution, given that also it is going to be extended for `mlcube.yaml` and `parameters.yaml` hash calculations as well.

cli/medperf/utils.py
hasan7n: alright. yes, there is actually no rationale behind having this. will do that change for `get_stats` as well.

aristizabal95 (approved):
Looks good! Thanks for going through the comments!

PR #284, opened 2022-10-18
MLCube read-only input parameters

### LINKED ISSUE #222: [FEATURE] Read-only permissions on cube inputs (singularity)
When using singularity-based cubes, we noticed that the inputs to these cubes are mounted with read-write permissions. This is unexpected for all cubes in a typical pipeline (data preparation, model and metrics) and could potentially cause reproducibility issues.

Possible solution:
In singularity, it is possible to specify mount options, which can be used to make a volume read-only. This should be the default for the input to cubes imho.

Note: This feature probably relies on a new feature of MLCube. Please raise issues as needed in that repo.

I haven't checked how docker-based cubes handle this.

aristizabal95:
This PR Implements read-only input parameters for mlcubes during the benchmark execution workflow. Closes #222.


hasan7n (changes requested):
Looks good!

Shouldn't we modify the test for `Result.create` as well? to ensure that the cubes also use read-only options.

hasan7n (approved):
great!

PR #286, opened 2022-10-19
fix server internal error during get_permissions

### LINKED ISSUE #280: [BUG] Benchmark Owner can't fetch individual results
**Describe the bug**
Benchmark owner should be able to access benchmark results, both from the `/benchmarks/{uid}/results` as well as `/results/{uid}`. Currently, the second option is failing, probably because of a bug with the isBenchmarkOwner check

**To Reproduce**
1. Run a server instance with benchmarks, models, datasets and results
2. Log in as the owner of a benchmark with existing results
3. Try to get results from the benchmark with `/results/{uid}` endpoint
4. See unauthorized error

**Expected behavior**
The endpoint should return individual results if the user is a benchmark owner / dataset owner related to that result

### LINKED ISSUE #287: [BUG] Retrieving non-existing entity results in internal server error
**Describe the bug**
Retrieving an entity (e.g. `GET /results/{uid}`) when the entity doesn't exists in the database returns a `500 internal error` response.

**To Reproduce**
call `GET /results/{uid}` with a non-existing result uid

**Expected behavior**
The response should carry a `404 not found` status

**Additional context**
1. Should also check endpoints other than `/results`
2. The problem seems to be due to this:
https://github.com/mlcommons/medperf/blob/06d42d0ab6c03200942fd1076108a70904692248/server/result/views.py#L52-L55
causing this error:
`TypeError: exceptions must derive from BaseException`

hasan7n:
This PR fixes a small bug during `GET /results/{uid}` permission checking, causing failure to retrieve individual results. Closes #280 

**Update:**
Closes #287 as well


aristizabal95:
Converting to Draft PR to make it clear it is not our intention to merge this yet

johnugeorge (approved):

PR #288, opened 2022-10-19
Mutable urls of hosted files

### LINKED ISSUE #194: URLs of hosted files should be mutable if hashes are stored separately
The purpose of disallowing changing file URLs stems from the need to ensure that a given experiment's code is immutable once the experiment has begun (scientific integrity).

However, in cases where the file hash is not part of the URL (e.g. not raw github content), this immutability is unnecessary and highly inconvenient. The file-hash suffices to ensure the code has not changed.

So any file URL that has an associated hash field should be mutable to enable changing file hosts mid-experiment if needed (e.g. a file server goes offline), since the URL plays no role in ensuring the code stays the same.

Note that for github raw user content links, the hash is part of the URL, thus those URLs must remain immutable.

### LINKED ISSUE #273: [FEATURE] MLCube files flexible hosting and hash checking
Currently, MLCube configuration files (`mlcube.yaml` and `parameters.yaml`)  are expected to be hosted on GitHub (MedPerf accepts only links starting with "https://raw.githubusercontent.com"). This was done to have a unique commit hash embedded in the link itself so that there is no need to compute the hash of the file and perform hash checking each time the files are downloaded.

Soon, we will support editing the links of any entity resource (see issue #194 ), while keeping the same hash. This will not be possible with the current MLCube files hosting requirements described above.

Just like `additional_files` link of the MLCube, configuration files should be treated the same. The files can be hosted anywhere, and extra server fields for MLCube should be created holding the hashes of these files. Hash computation and checking for these files should then be performed by the client.

hasan7n:
This PR allows URLs of the following hosted files to be editable:
- demo dataset url
- mlcube parameters file
- mlcube manifest file
- mlcube additional files tarball
- mlcube container image tarball file

Closes #194 

Additionally, this PR will remove the constraint of hosting the mlcube manifest and parameters files on GitHub. This is to allow these URLs to be mutable as well. Closes #273


server/mlcube/migrations/0005_auto_20221101_1524.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.16 on 2022-11-01 15:24
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('mlcube', '0004_generate_missing_hashes'),
```
johnugeorge, line 9: Which is this migration?

server/mlcube/migrations/0004_generate_missing_hashes.py
```diff
@@ -0,0 +1,52 @@
...
+
+    dependencies = [
+        ("mlcube", "0003_auto_20221021_2038"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            create_missing_hashes, reverse_code=migrations.RunPython.noop
+        ),
+    ]
```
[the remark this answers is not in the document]
johnugeorge: My only concern is that we are adding cli logic to server api. For eg: what if hash calculation logic changes in CLI? Will server logic remain consistent?

aristizabal95 (commented):
Everything looks good on the client side! Waiting on Johnu's opinion

server/mlcube/models.py
```diff
@@ -10,7 +10,9 @@ class MlCube(models.Model):
 
     name = models.CharField(max_length=20, unique=True)
     git_mlcube_url = models.CharField(max_length=256)
-    git_parameters_url = models.CharField(max_length=256)
+    mlcube_hash = models.CharField(max_length=100)
```
johnugeorge, line 13: I see two migrations: Is it because you tried changing the field attributes in two steps?

johnugeorge:
I see that editable fields in operation has changed. Please verify from your side as well. Also, please rebase your changes as the branch is out-of-date

server/mlcube/models.py
hasan7n: True. I will merge them

hasan7n:
@johnugeorge Yes, as pointed out by the PR's description, URLs can be edited anytime given the fact that hash integrity is checked every time at the client.

johnugeorge (approved):

PR #292, opened 2022-10-21
Check cube validity on download. Redownload if failed

### LINKED ISSUE #239: [FEATURE] Cube Checksum and redownload
One of the biggest issues collaborators encountered during the FeTS Challenge was network issues during model retrieval. This usually caused failed image downloads that would then be either non-existent or corrupted. In these situations, we had to ask the collaborator to manually remove the failing files and try again.

Medperf should be able to identify these situations, and retry model retrieval for a maximum number of times before reporting the issue. This is usually done with a checksum, which is something we already have implemented. This can also be implemented for additional files

aristizabal95:
This PR adds a quality-of-life feature, which automates an annoying process of removing and redownloading failing cubes, which are corrupted due to failed download attempts. 

This is fixed by checking cube validity when using `get`, and if validation fails, removes the cube and downloads again up to a maximum number of times.

Closes #239.


cli/medperf/entities/cube.py
```diff
@@ -122,9 +123,18 @@ def get(cls, cube_uid: str) -> "Cube":
...
+            logging.info(f"Downloading MLCube. Attempt {attempt + 1}")
+            cube.download()
+            if cube.is_valid():
+                cube.write()
+                return cube
+            attempt += 1
+        logging.error("Max download attempts reached")
+        cube_path = os.path.join(storage_path(config.cubes_storage), str(cube_uid + 1))
+        cleanup([cube_path])
+        raise RuntimeError("Could not successfully download the requested MLCube")
```
hasan7n, line 137: I think we should keep medperf failures exiting with pretty errors rather than python exceptions. What do you think. (exception --> we have an internal bug)

cli/medperf/tests/entities/test_cube.py
```diff
@@ -340,10 +353,68 @@ def test_get_cube_with_image_untars_image(mocker, comms, img_body, no_local):
...
+    mocker.patch(PATCH_CUBE.format("Cube.is_valid"), return_value=False)
+    spy = mocker.patch(PATCH_CUBE.format("cleanup"))
+    cube_path = os.path.join(storage_path(config.cubes_storage), str(uid))
+
+    # Act
+    with pytest.raises(RuntimeError):
+        Cube.get(uid)
+
+        # Assert
+        spy.assert_called_once_with(cube_path)
```
hasan7n, line 399: I was wondering why this test was not failing :) It turns out that the `spy.assert...` should be called outside the `pytest.raises` block. Same thing for the test above this one.

hasan7n (changes requested):
Thanks Alejandro! I left a few comments.

cli/medperf/entities/cube.py
```diff
@@ -122,9 +123,18 @@ def get(cls, cube_uid: str) -> "Cube":
...
+        attempt = 0
+        while attempt < config.cube_get_max_attempts:
+            logging.info(f"Downloading MLCube. Attempt {attempt + 1}")
+            cube.download()
+            if cube.is_valid():
+                cube.write()
+                return cube
+            attempt += 1
+        logging.error("Max download attempts reached")
+        cube_path = os.path.join(storage_path(config.cubes_storage), str(cube_uid + 1))
```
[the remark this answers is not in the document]
aristizabal95: I have no idea what was going through my brain when I did this! Oops thanks for noticing it!

cli/medperf/entities/cube.py
aristizabal95: For now we can do that, but following from a previous discussion on another PR, I think entities should not call the UI directly. Using `pretty_error` restricts future implementations that might want to use medperf's entities and API from deciding when to interact with the UI and how to handle errors. What I think is the best approach for a future PR is to move all calls to `pretty_error` to the commands layer, and instead create our own exceptions that are called inside entities and utils. This way we're separating concerns, since entities no longer have to be aware of the UI.

cli/medperf/tests/entities/test_cube.py
aristizabal95: Interesting! I wasn't aware of this! I'll make the required changes

cli/medperf/entities/cube.py
aristizabal95: Let's talk about this in our discussion. I do agree that I should follow the way we're doing things now until we create a PR for custom exceptions.

hasan7n (approved):
Looks great thanks Alejandro!

PR #293, opened 2022-10-24
fix the failing mkdocs command

hasan7n:
A simple change (removal of an extra whitespace). It was causing `mkdocs` command to fail when building the docs.


aristizabal95 (approved):
Very simple indeed!

PR #300, opened 2022-10-25
Use pagination for APIs that return lists

### LINKED ISSUE #256: [BUG] Server errors out on responses with body size greater than 32M
**Describe the bug**
Some endpoints are expected to keep growing in response size following platform usage. These endpoints will most probably return responses with more than 32M of size (as has happened with the results for the FeTS Challenge), which gets blocked by GCP, raising a 500 Internal Error.

**To Reproduce**
1. Run a medperf instance through GCP
2. Post multiple objects of the same type (datasets for example), where in total, the contents of all entities goes beyond 32M in size
3. get all previously generated entities with the bulk get endpoint (GET /datasets/ for example).
4. The server will fail and return 500 Internal Error

**Expected behavior**
The server should either allow for larger responses, or provide tools to reduce the size of the response through something like pagination. In cases where the size surpasses 32M, a meaningful error description should be provided to the user.

**Screenshots**
Screen Shot 2022-09-19 at 12 54 16 PM
Failed GET request because the FeTS Challenge has over 730 results, each of which is large by itself.

**Tasks**

- [ ] #294
- [x] #295

### LINKED ISSUE #294: Enable pagination on all GET list calls

### LINKED ISSUE #295: Implement logic for retrieving all pages, and falling back when an error is encountered

hasan7n:
This PR modifies the response of list APIs (e.g. `GET /me/mlcubes`) to allow returning paginated responses, rather returning all results in one response.

Before:
`GET /me/mlcubes` --> [`cube1`, `cube2`, `cube3`]

Now:
`GET /me/mlcubes?limit=2&offset=0` -->
```
{
        "count": 3,
        "next": `/me/mlcubes?limit=2&offset=2`,
        "previous": None,
        "results":  [`cube1`, `cube2`]
 }
```
Closes #294 .
This PR will additionally modify the cli to cope with the server changes, thereby closes #295 and consequently closes #256


aristizabal95:
Given that we use the name `results` in other places, I believe a better name for this field would be `payload` to reduce confusion.

hasan7n:
@aristizabal95 I added logging so that errors are logged even if they were due to page size. I think it is useful for debugging. I am a bit hesitant here to keep it like this or log warnings instead.

cli/medperf/comms/rest.py
```diff
@@ -114,6 +114,61 @@ def __req(self, url, req_func, **kwargs):
                 "remember to provide the server certificate through --certificate"
             )
 
+    def __get_list(
+        self,
+        url,
+        num_elements=None,
+        page_size=config.default_page_size,
+        offset=0,
+        binary_reduction=True,
```
johnugeorge, line 123: Do we need to keep it as True? From one of our earlier discussions, return-code != 200 can happen because of reasons other than the page limit issue

aristizabal95: Changed the default to False. After the discussion I understand the concerns with defaulting to True, and I can see this should only be used on extraordinary cases where we know dynamically changing the size of the page will help overcome an issue.

johnugeorge (approved):

PR #302, opened 2022-10-25
Process-dependent temporary storage

### LINKED ISSUE #277: [BUG] Cleanup might affect concurrent commands
**Describe the bug**/**To Reproduce**
If we are expecting users to run simultaneously multiple medperf commands, we should consider the following scenario:

command A: a certain benchmark execution
command B: preparation of some dataset

commands A and B are being run simultaneously.

if command A fails for some reason (or finished too early), cleanup will be called by default, erasing all folders starting with tmp_, probably including a folder currently being using by command B.

**Solution**

@aristizabal95's elegant suggestion (it is cool): We can get the python execution PID as described [here](https://www.geeksforgeeks.org/python-os-getpid-method/) and append it to the `tmp_` suffix at the beginning of every execution. This way, when we reach the cleanup procedure, it will by default only clean up things that are named with `tmp_{PID}_`.

aristizabal95:
This PR implements process-dependent temporary storage. Having process-dependent temporary storage allows for simultaneous execution of commands, avoiding cleanup collision. This is done by appending the process ID to cleanup-related files at the beginning of every command. Closes #277.

### DIFF
--- a/cli/medperf/__main__.py
+++ b/cli/medperf/__main__.py
@@ -14,7 +14,7 @@ import medperf.commands.mlcube.mlcube as mlcube
 import medperf.commands.dataset.dataset as dataset
 from medperf.commands.auth import Login, PasswordChange
 import medperf.commands.benchmark.benchmark as benchmark
-from medperf.utils import init_storage, storage_path, cleanup
+from medperf.utils import init_storage, storage_path, cleanup, set_unique_tmp_config
 import medperf.commands.association.association as association
 from medperf.commands.compatibility_test import CompatibilityTestExecution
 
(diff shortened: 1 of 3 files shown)


aristizabal95:
Had to add a cleanup procedure to revert the config singleton to its original state. I honestly don't like this, so any suggestion is highly appreciated

hasan7n (commented):
Look great Alejandro! I really like this pid stuff.

Just questions before approving:
1. shouldn't we also append the pid to `cube_submission_uid` ? it is removed also during cleanup.
2. do you think medperf logs may become cumbersome during simultaneous processes? I think this is out of this PR's scope but just mentioning.

also I edited the test hope it looks good :)
**Edit:**
No, the test doesn't actually check what we want. Should revert the commit.
Any thoughts about mocking a module?

hasan7n (approved):

PR #303, opened 2022-10-25
Mkdocs build fix and install instructions

aristizabal95:
This PR addresses a hidden bug that caused mkdocs to fail at build/serve time. This was caused by a build param that was updated some time ago.

Additionally, this PR includes a `requirements.txt` file and instructions for running mkdocs locally, for documentation contributions.

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.x
-      - run: pip install mkdocs-material mkdocstrings[python] mkdocs-gen-files mkdocs-literate-nav
+      - run: pip install -r docs/requirements.txt
       - run: git config --global user.email "github-actions[bot]@users.noreply.github.com"
       - run: git config --global user.name "github-actions[bot]"
       - run: mkdocs gh-deploy --force
--- a/README.md
+++ b/README.md
@@ -196,3 +196,15 @@ Once all the requirements are met, running `sh test.sh` will:
 - Retrieve the demo dataset
 - Run the CLI demo using `cli/cli.sh`
 - cleanup temporary files
+
+## Documentation Contribution
+If you wish to contribute to our documentation, here are the steps for successfully building and serving documentation locally:
+
+- **Install dependencies:** We are currently using mkdocs for serving documentation. You can install all the requirements by running the following command:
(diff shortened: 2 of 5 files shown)


hasan7n (changes requested):
It's great having the instructions!

**In brief:**
github workflow building the docs should fail now. We should revert back the change in `gen_ref_pages.py` and 'correctly' install mkdocs requirements, which are listed in `docs-ci.yml`, to have successful local serving. Let me know if I am missing something.

**Long answer:**

https://github.com/mlcommons/medperf/blob/a39b84143cdcc39016a7f12ce30b82c9687f4210/docs/gen_ref_pages.py#L12

Regarding the above change, I remember I did that for the GitHub workflow `.github/docs-ci.yml`. After this PR, local serving works but github workflow will fail (tested on my fork). Can you confirm?

It seems that this is a problem of requirements. If we locally install the following as stated in `docs-ci.yml`:
`pip install mkdocs-material mkdocstrings[python] mkdocs-gen-files mkdocs-literate-nav`
Local serving will work fine.

So we have two options:
1. update docs-ci.yml to install the listed requirements in this PR
2. revert back the change in `gen_ref_pages.py`, and modify the requirements.txt to contain what the current docs-ci.yml contains.

I would go with 2. since I noticed locally that latest versions of mkdocs libraries (not sure which one) give rich warnings when docstrings don't resemble function signatures.


+ to remember: going with 1. requires a small change to hide tests from documentation

Let me know what you think. I can do the changes but want you to confirm-test if local serving will work if we used only the requirements listed in `docs-ci.yml` (which works for me in pythons 3.7 and 3.10).

hasan7n:
by the way, do you think it would be good if we add a new check for pull requests to test if `mkdocs build` works?

aristizabal95:
Hey Hasan, thanks for the explanation! I was sure there was a reason behind the `cli/medperf` that I was removing. I totally agree that we should stick with the steps provided within `docs-ci.yml`. I tried doing the local installation with those packages, but I'm not able to run `mkdocs serve`. `jinja2==3.0.0` seems to be required, but even then I get errors like `ERROR    -  Could not collect 'config'` which I don't get when using the `requirements.txt` provided on this PR. I think we should look into this to discover the discrepancy, so we can modify this PR accordingly.

hasan7n (approved):
It works both locally and on Github!

PR #309, opened 2022-11-04
Client configuration profiles

### LINKED ISSUE #197: [FEATURE] Configuration Management
The configuration is being managed through a `config.py` file, which needs to be manually edited to override the defaults.

The configuration values should be stored in a .yaml file, which is then read and processed by the current config.py.
This config.yaml file would be copied to the `~/.medperf` on first execution.
Users should be able to change the default values via a command (`medperf config set [KWARGS]`)
Users can reset the configuration to the default settings via a command (`medperf config reset`)

aristizabal95:
This PR adds logic to handle multiple configuration and authentication profiles. It additionally separates the filesystem by server, so that resources are shared for the same server. Lastly, commands for managing profiles are provided, so users can easily create and adjust their settings.

Closes #197.


cli/medperf/commands/profile.py
```diff
@@ -0,0 +1,117 @@
...
+    Arguments:
+
+    {0}
+    """
+    args = parse_context_args(ctx.args)
+    config_p = read_config()
+
+    filtered_params = {k:v for k, v in args.items() if k in config.customizable_params}
+    if name in config_p:
+        pretty_error("A profile with the same name already exists", config.ui)
```
hasan7n, line 45: Extra ui arg

cli/medperf/commands/profile.py
```diff
@@ -0,0 +1,117 @@
...
+    config_p[name] = filtered_params
+    write_config(config_p)
+
+
+@app.command("set", context_settings={"allow_extra_args": True, "ignore_unknown_options": True})
+@docstring_parameter(" | ".join(config.customizable_params))
+def set(ctx: typer.Context):
+    """Assign key-value configuration pairs to the current profile.
+
+    Extra arguments in the format "--key=value" will be handled as custom configuration
```
hasan7n, line 56: just remove the "Extra". I got lost for a while :)

cli/medperf/commands/profile.py
```diff
@@ -0,0 +1,117 @@
...
+    Args:
+        profile (str, optional): Profile to display information from. Defaults to active profile.
+    """
+    ui = config.ui
+    if profile is None:
+        profile = config.profile
+
+    config_p = read_config()
+    profile_config = config_p[profile]
+    dict_pretty_print(profile_config, ui)
```
hasan7n, line 117: extra ui :)

cli/medperf/commands/profile.py
```diff
@@ -0,0 +1,117 @@
...
+    Available arguments:
+
+    {0}
+    """
+    profile = config.profile
+    args = parse_context_args(ctx.args)
+    config_p = read_config()
+
+    filtered_params = {k:v for k, v in args.items() if k in config.customizable_params}
+    current_config = config_p[profile]
```
hasan7n, line 67: I am concerned about this.
A small typo in the key name (e.g. timeout->timeuot) when configuring a profile would have unexpected consequences.

cli/medperf/utils.py
```diff
@@ -21,9 +22,69 @@
...
+
+    Args:
+        ctx_args (List[str]): List of extra cli arguments
+
+    Returns:
+        dict: dictionary of key-value cli arguments
+    """
+    args = []
+    for arg in ctx_args:
+        args += arg.split("=")
```
hasan7n, line 36: currently this will work:
`medperf profile set --key=val=--=`
Let's have here an assertion that only one `=` sign is present. (idealism :))

cli/medperf/comms/rest.py
```diff
@@ -75,14 +76,19 @@ def change_password(self, pwd: str) -> bool:
...
+        if os.path.exists(creds_path):
+            creds = configparser.ConfigParser()
+            creds.read(creds_path)
+            if profile in creds:
+                self.token = creds[profile]["token"]
+                return
+
+        pretty_error(
+            "Couldn't authenticate. Did you run 'medperf login' before?",
+            self.ui,
```
hasan7n, line 90: extra ui :)

cli/medperf/__main__.py
```diff
@@ -123,53 +134,26 @@ def test(
...
-    evaluate_timeout: int = config.evaluate_timeout,
-    platform: str = config.platform,
-    cleanup: bool = True,
-    certificate: str = config.certificate,
-    local: bool = typer.Option(
-        False, help="Run the CLI with local server configuration"
-    ),
+    profile: str = typer.Option(
+        "default", help="Configuration profile to use"
+    )
```
hasan7n, line 139: Currently, if a user wants to keep using their custom profile, they should keep using `--profile mycustomprofile` with every command.
I suggest the following:
- In addition to the `default` profile, we keep a reserved profile called `active`.
- We add a command: `medperf profile activate <profile_name>`. This would copy the specified profile into the reserved `active` profile.
- medperf always loads the `active` profile.

cli/medperf/utils.py
hasan7n: I just realized: should be only one `=` sign or no `=` signs.

hasan7n (changes requested):
Awesome work Alejandro! It's not an easy issue!
Here is a summary of my detailed comments:
1) Some very minor stuff (extra `ui`s,  docstring edit)
2) Invalid params for `profile` should raise errors rather than be ignored. 
3) `parse_context_args` modification.
4) have an `active` profile and a `medperf profile activate` command.

Additionally, I would like to suggest the following, however, it might be just a perspective :)
Having an "eager callback" to dynamically set the Typer options as described [here](https://github.com/tiangolo/typer/issues/86#issuecomment-996374166) can be used instead of `Typer context`. Would save us from parsing context and checking for invalid args.

cli/medperf/commands/profile.py
aristizabal95: You are right, instead of filtering, we should validate and throw errors if an invalid argument is passed. I'll implement this

cli/medperf/__main__.py
aristizabal95: I had implemented something similar to this, but we encounter an issue of desyncing profiles. Say I run `medperf profile active fets`. I will assume that whatever changes I do to that profile will be reflected to the active profile and viceversa. That will not be the case.

I'll see if I can make this active profile just store the name of the active profile and use that to assign the default profile.

aristizabal95:
I made some modifications which I believe make the code more readable and less hacky. Configuration options now are managed through a decorator. Using the `merge_args` library, it was possible to attach new options to existing commands in a way that typer can handle directly. We no longer need to parse arguments directly, nor format help messages through unconventional means.

cli/medperf/commands/profile.py
```diff
@@ -0,0 +1,111 @@
...
+    args = ctx.config_dict
+    config_p = read_config()
+
+    current_config = config_p[profile]
+    current_config.update(args)
+    config_p[profile] = current_config
+    write_config(config_p)
+
+
+@app.command("unset")
```
hasan7n, line 75: Suggestion: remove or rename the unset command. The users don't actually think about it as 'setting and unsetting'; they are just changing parameters. `unset` actually `reverts to default`, but it appears like you are 'removing' a config param.
We can change its name (reset, revert), or just remove the command. A user that uses `unset` should be aware of what is the default they are reverting to.. They can simply use `set` again. Let me know what you prefer.

I usually refer to how `gcloud` handles this stuff. Looking at https://cloud.google.com/sdk/gcloud/reference/config/unset
The unset actually makes the parameter empty, not resetting it to a default value. They have a notion of an empty parameter that, when needed, the command fails and asks the user to set it.

cli/medperf/commands/profile.py
```diff
@@ -0,0 +1,111 @@
...
+    config_p = configparser.ConfigParser()
+    config_path = os.path.join(config.storage, config.config_path)
+    config_p.read(config_path)
+    return config_p
+
+
+def write_config(config_p: configparser.ConfigParser):
+    config_path = os.path.join(config.storage, config.config_path)
+    with open(config_path, "w") as f:
+        config_p.write(f)
```
hasan7n, line 23: they can be moved to utils, however, keeping them doesn't hurt :)

cli/medperf/utils.py
```diff
@@ -22,9 +23,42 @@
...
+    config.profile = profile
+    return config_p[profile]
+
+
 def storage_path(subpath: str):
     """Helper function that converts a path to storage-related path"""
-    return os.path.join(config.storage, subpath)
+    server_path = config.server.split('//')[1]
+    server_path = re.sub(r"[.:]", "_", server_path)
+    return os.path.join(config.storage, server_path, subpath)
```
hasan7n, line 61: 1) We need to keep in mind then that we **should** use `storage_path` **and not** `config.storage` for everything in the code, except credentials and config. I double-checked, and we are good.

2) Just wondering: does it need an `os.makedirs(exist_ok=True)` for the first time the server path is created?

hasan7n (changes requested):
Hey @aristizabal95 ! Looks great now! I left a few comments.

cli/medperf/commands/auth/login.py
```diff
@@ -1,8 +1,8 @@
 import os
 import stat
+import configparser
 
 import medperf.config as config
-from medperf.utils import storage_path
 
 
 class Login:
```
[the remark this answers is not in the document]
hasan7n: I wonder though how the client integration test check passes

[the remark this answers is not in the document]
aristizabal95: I believe this is happening because you still have a `credentials` file from previous versions, which stored things differently. This is a breaking change, so it's best to try it on a clean medperf environment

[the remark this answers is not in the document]
hasan7n: True :)

cli/medperf/decorators.py
```diff
@@ -27,3 +31,90 @@ def wrapper(*args, **kwargs):
...
+                "--platform",
+                help="Platform to use for MLCube. [docker | singularity]",
+            ),
+            cleanup: bool = typer.Option(
+                default_val,
+                cleanup_opts,
+                help="Wether to clean up temporary medperf storage after execution",
+            ),
+            **kwargs,
+        ):
```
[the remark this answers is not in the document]
aristizabal95: Looking back at it, it took me a while to understand why that line of code is there. `assigned_params` catches all the options passed to the main command, which may not be desired to be written to the config file. `kwargs` catches the options specified to main. A small example might explain this better

Running
`medperf --no-cleanup profile set --platform=docker`
Implies a few things (These assumptions can be discussed of course):
- options passed before the subcommand are temporary inline options. `--no-cleanup` in this case should only be used for this specific command
- options passed after the subcommand are desired to be processed by the specified subcommand. `--platform=docker` should be used by the `profile set` command to store it.
That's what the `set(assigned_params) - set(kwargs.keys())` is able to differentiate.

For main, we also have the `--profile` option which is specified to `active` by default. Without that substraction, all profile configurations would end up with a `profile=active` stored on their configuration dictionary.

I will add a command that (hopefully) explains why that line is there briefly.

cli/medperf/decorators.py
```diff
@@ -27,3 +31,90 @@ def wrapper(*args, **kwargs):
...
+        with_args (bool, optional): Wether to expect arguments or just flags. Defaults to True.
+
+    Returns:
+        Callable: decorated function
+    """
+
+    def decorator(func):
+        atype = str
+        cleanup_opts = "--cleanup/--no-cleanup"
+        default_val = None
```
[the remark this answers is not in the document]
aristizabal95: That's true. This was done this way so that we could differentiate assigned vs unassigned config values. I had a thought about keeping configuration profiles slim by only storing the difference from the default configuration, but it might be better to have defaults and always store all configuration variables for each profile. I believe Micah also said that would be more intuitive.

I'll make the required change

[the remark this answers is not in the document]
aristizabal95: Hmm will have to look more into this, because providing the default values makes the command think we specified all parameters, which might lead to undesired behavior. For example, running `medperf profile set` will set all parameters to the default values. Additionally, running any medperf command will think we passed all default values as inline cli arguments, invalidating any profile configuration. We can do some workarounds, but in those cases we the `--help` text will have to say None

cli/medperf/utils.py
aristizabal95: 1. That's true. So far, we only keep track of two things in the global storage path (config and credentials). If we start making more use of that folder we should probably create another helper function for that (probably `global_storage_path`?) For now I don't think we need it

2. No need, this is covered by the `init_storage` function, which is always executed before anything else

cli/medperf/decorators.py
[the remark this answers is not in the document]
hasan7n: I see. thanks for the explanation!

cli/medperf/decorators.py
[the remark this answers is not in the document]
hasan7n: I see your point. I think we should move on with this PR although this "default=None" issue can still be a concern.
A workaround can be to have the `configurable` decorator load the profile and sets the defaults of that specific profile that is being used. for the `main` command where the decorator runs before config initialization, we can move the config initialization logic to happen in the decorator. i.e. have something like @configurable(init_config=True) for the `main` command.

However, this really complicates stuff and adds a lot of logic and an extra argument to the decorator we have, which is already complex.
I think we can move on and merge this PR and create an issue to resolve this later, perhaps to look more into how this type of stuff is handled. So basically we create an issue to resolve later that addresses two confusions:
1) User confusion when they type` --help` and see that the defaults are `None`
2) developer confusion about the `kwargs` in the decorator.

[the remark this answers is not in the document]
aristizabal95: Agreed! It's best to move forward and have those concerns as issues so that we can move forward

PR #310, opened 2022-11-04
Option for allowing to submit partial results

### LINKED ISSUE #237: [FEATURE] Option for allowing execution of the benchmark pipeline even if one of the mlcubes fail
Currently, if an MLCube fails, medperf stops execution and prevent any future steps to be executed. This may not be desirable in some specific scenarios. To give a concrete example, one of the FeTS models failed during inference time, but it was able to generate predictions for a portion of the input data. Generating results for these predictions is desirable, to take them into account during the evaluation phase. Thus, having an easy way to allow complete pipeline execution would be useful.

Having some option like `--force-execution` or `--ignore-errors` which would basically disable the exit-status check inside `Cube.run` should suffice.

hasan7n:
This PR adds `--ignore-errors` option for commands `medperf run` and `medperf result create`. When True, errors will be ignored if mlcubes fail and `Result` creation proceed successfully. Also, the `metadata` field sent to the server is updated in case of ignored errors.

Closes #237


cli/medperf/commands/result/create.py
```diff
@@ -50,6 +60,8 @@ def __init__(
         self.evaluator = None
         self.model_cube = None
         self.run_test = run_test
+        self.ignore_errors = ignore_errors
+        self.metadata = {"partial": False}
```
aristizabal95, line 64: Awesome addition!

aristizabal95 (approved):
Looks great! I really like how you use the metadata to inform about wether results are partial or not!

PR #311, opened 2022-11-05
Check the existence of compatibility test results

### LINKED ISSUE #248: [FEATURE] Cache and check existence of compatibility test results
Compatibility tests can take a very long time to finish depending on the use-case. This can be annoying and problematic in some cases, given that these tests are required for association requests. Even when an association compatibility test is completed, the association may be cancelled because of wrong input, or terminal timeout. Currently, if this happens, the previous test is ignored, and another test is initiated which could cause high delays and wasted computation.

Test results should be cached in such a way that they are uniquely identified, and easily found in subsequent test executions. If a test is initiated, but results for that test already exist, then load those results and skip the test execution.

**Describe alternatives you've considered**
We have provided an autoapprove option that allows the user to avoid scenarios where the approval step might go wrong right at the end. This still won't help in cases where the test is executed twice.

**Additional context**
This is an issue that one of the FeTS contributor has been dealing with, and it has been a very annoying process of executing the compatibility test multiple times only for it to fail due to timeout in the end. The autoapprove argument was provided to avoid the same situation, but a better fix must be given.

hasan7n:
This PR allows compatibility tests to check for existing results before getting executed. If results exist, then the test will not run, unless the `force-test` flag is set. Closes #248 .


cli/medperf/commands/compatibility_test.py
```diff
@@ -220,3 +225,20 @@ def download_demo_data(self):
         data_path = os.path.join(untar_path, paths["data_path"])
         labels_path = os.path.join(untar_path, paths["labels_path"])
         return data_path, labels_path
+
+    def cached_result(self):
```
aristizabal95, line 229: A docstring here explaining the intention of the function, and in which cases it returns None would be useful here

aristizabal95 (changes requested):
Hey Hasan, this looks good! I'd only ask you to add a docstring to the cache function. After that we can merge this!

aristizabal95 (approved):
Awesome Hasan! Thanks a lot for this!

PR #315, opened 2022-11-09
Decouple UI from low-level logic

### LINKED ISSUE #297: [FEATURE] Custom Exceptions for entities
Currently, when entities encounter situations that should cause errors, they are calling `pretty_print` to display a message to the user. This is not the best practice, as we're forcing entities to be aware of (and hardcoding interaction with) the UI. 

Entities should instead throw meaningful exceptions that are then handled by the next layer (in this case `commands`).

aristizabal95:
This PR removes calls that interact with the UI from low-level functionality, like entities and communication layers.  Direct calls to UI print funcionality have been removed, given they don't provide too much value (like indicating when a hash is generated, which the user doesn't really care about) or they can be handled on a different manner (like printing errors, which can be better managed through custom exceptions and decorators for pretty printing such scenarios).

With this, low-level functionality isn't concerned with how to display information to the user, which allows for greater control and usability of our code.

Closes #297.

### DIFF
--- a/cli/medperf/comms/factory.py
+++ b/cli/medperf/comms/factory.py
@@ -1,6 +1,6 @@
 from .rest import REST
 from .interface import Comms
-from medperf.utils import pretty_error
+from medperf.exceptions import InvalidArgumentError
 
 
 class CommsFactory:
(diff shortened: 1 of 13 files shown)


hasan7n (approved):
Amazing! looks great now.
A note: after merging this PR, let's immediately modify the current PRs since some pretty errors might stay silently (came to my mind the pagination PR. There is a pretty error in comms)

PR #317, opened 2022-11-10
Update local entities by default

### LINKED ISSUE #298: [FEATURE] Sync entities that are going to be used by a command
 Local entities can get out-of-sync at any time, due to changes done either locally or remotely. Examples are associated models to a benchmark, mlcube modifiable fields, or approval status. These changes should be reflected locally.

When executing a command that makes use of a specific entity, try to sync it with the server first before doing anything else.
(e.g. medperf run -b 1 -m 2 -d 63a --> update cache for benchmark 1, model 2 and dataset accordingly)

aristizabal95:
This PR adds the required functionality for entities to stay up-to-date with the server by defaulting to retrieving information from the server. If an error is encountered during remote data retrieval, then the client falls back to looking for local cache.

For MLCubes, the metadata is always downloaded first, and then validity of the cube is checked in case all the related heavy assets have already been downloaded. Because of this, many unrelated tests had to be updated to allow the execution of the download procedure.

Closes #298.


cli/medperf/entities/benchmark.py
```diff
@@ -70,31 +71,33 @@ def all(cls) -> List["Benchmark"]:
...
             add_models = cls.get_models_uids(benchmark_uid)
             benchmark_dict["models"] = [ref_model] + add_models
+        except CommunicationRetrievalError:
+            # Get local benchmarks
+            logging.warning(f"Getting benchmark {benchmark_uid} from comms failed")
+            logging.info(f"Looking for benchmark {benchmark_uid} locally")
+            bmk_storage = storage_path(config.benchmarks_storage)
+            local_bmks = os.listdir(bmk_storage)
+            if str(benchmark_uid) in local_bmks:
+                benchmark_dict = cls.__get_local_dict(benchmark_uid)
```
hasan7n, line 100: missing an else statement to raise an error about not finding the entity neither locally nor in the server.
Now this scenario would raise a NameError

cli/medperf/entities/cube.py
```diff
@@ -115,30 +115,38 @@ def get(cls, cube_uid: str) -> "Cube":
...
-            if cube.is_valid():
-                cube.write()
-                return cube
-            attempt += 1
-        logging.error("Max download attempts reached")
+
+        # Try to download the cube first
+        try:
+            meta = comms.get_cube_metadata(cube_uid)
+            cube = cls(meta)
```
hasan7n, line 124: missing a `cube.download()` after this line. Currently, `cube.is_valid()` will always return False at the first attempt.

cli/medperf/entities/benchmark.py
hasan7n: same comment for `result` and `dataset` entities.

hasan7n (changes requested):
Awesome! Please find the comments above. Minor stuff

cli/medperf/entities/cube.py
aristizabal95: This is by design. `cube.is_valid()` will fail if the required files are not present (which would be the case in a first-run). But, if the cube has already been downloaded before, we don't want to force to re-download images and additional files which are expensive. By checking validity first, we ensure already-downloaded skip the download procedure

cli/medperf/entities/benchmark.py
aristizabal95: Will do!

cli/medperf/entities/cube.py
hasan7n: oh right that's perfect!

hasan7n (approved):
Super!

PR #318, opened 2022-11-13
A command to run all models for a benchmark

### LINKED ISSUE #207: [FEATURE] Command to run all outstanding models in a benchmark
A data owner should be able to run a command on a given dataset and benchmark that:
1. Determines all models included in a benchmark
2. Determines the sub-list of models without current results for that dataset
3. For each model that lacks results for that dataset, in priority order (as defined by the association), creates the results
4. Results are not approved/uploaded.

### LINKED ISSUE #351: [BUG] Errors of metrics cube should not be ignored
about the ignore_errors option for running a benchmark experiment:
It ignores the mlcube failure and hence the client would upload partial results.
The current logic ignores the errors of both the model cube and the metrics cube, which sounds to not make sense. A failed metrics mlcube will not generate the results file at all.

### LINKED ISSUE #364: [BUG] result ls doesn't work if a benchmark-submission result exists
The error displayed:
`❌ The requested dataset information could not be found locally.`

**Reason**
The following happens when listing results: the `Result` constructor is called, and it retrieves the dataset associated with this result to construct the result's generated uid.
The following happens when submitting a benchmark:  Benchmark submission generates a result, and cleanup is called after this command is finished. Cleanup deletes the dataset associated with the generated result.

**Solution**
The solution should be straightforward. However, this issue raises some questions:
1) Why are we deleting prepared demo datasets?
2) Should we keep differentiating demo datasets from other datasets by the `test_` prefix?
3) Should we find a better convention about the result generated uid?

hasan7n:
This PR implements a new command `medperf benchmark run` that runs all outstanding models of a given benchmark on a given. Closes #207 . Closes #351.

This PR is still in draft mode since it seems it depends on two major changes to be worked on after discussion with the team:
1. Raise exceptions everywhere instead of using `pretty_error` (use `pretty_error` only at the end of `medperf` process, inside the decorator).
2. The logic for listing all entities from the server and the local storage currently resides in `command.list` but with no way to reuse it. This logic is needed for the new command of this PR.

---
  - https://app.asana.com/0/0/1203887043177430


hasan7n:
TODO: `BenchmarkSubmission` should use `BenchmarkExecution`, not `CompatibilityTestExecution`

cli/medperf/commands/benchmark/run.py
```diff
@@ -0,0 +1,129 @@
...
+        for model in self.outstanding_models:
+            exec_summary = {
+                "model": model,
+                "success": False,
+                "partial": "",
+                "error": "",
+            }
+            # reset
+            self.model_execution.model_uid = model
+            self.model_execution.out_path = results_path(
```
aristizabal95, line 74: This is no longer supported. `results_path` is no longer an utils function. This also makes me wonder, can't we just create new `BenchmarkExecution`s for each model, and execute the `run` method here?

cli/medperf/commands/benchmark/run.py
```diff
@@ -0,0 +1,129 @@
...
+                    f"There was an error when executing the benchmark with the model {model}: {e}"
+                )
+                exec_summary["error"] = str(e)
+                self.summary["executions"].append(exec_summary)
+                continue
+
+            self.model_execution.write()
+            self.model_execution.remove_temp_results()
+
+            exec_summary["success"] = True
```
aristizabal95, line 103: I like the idea of handling errors and providing a summary of the complete execution here. Nice!

aristizabal95 (changes requested):
Hey Hasan, this looks great! Providing an execution summary is a great idea! 

My main concern is that we're explicitly specifying the benchmark execution workflow on this new command, instead of using the workflow provided by the `BenchmarkExecution` command. This could cause issues in the future, if we modify the `BenchmarkExecution` workflow, because it could cause deviations on how benchmarks are executed if done individually vs by bulk. My suggestion would be that once we know which models to execute, run the `BenchmarkExecution.run` function directly. 

Let me know if you have some thoughts on this.

hasan7n:
@aristizabal95 I totally agree.
My concern was if we create a new benchmark execution for each model, let us say 20 models, this will retrieve the same metrics cube 20 times, as well as retrieve the dataset metadata 20 times. This is because we are planning to update entities when we want to use them.

We can do as above, but it will be very inefficient. I was having a thought of generalizing the current BenchmarkExecution class so that it can itself be used for 1 or more models. What do you think about this? It sounds good to me.

hasan7n:
So basically we modify BenchmarkExecution class to work with a list of model cubes. If the class receives a model cube, the list of cubes will then be only one model cube. If the class didn't receive a model cube, it will do the logic of retrieving the outstanding models.
But of course, we should keep these commands separate in the user interface.
i.e.
benchmark run -b -m
benchmark run_all -b

but both commands use the same generalized class

aristizabal95:
Sounds good! Hopefully it should not require a lot of changes!

hasan7n:
TODO: add integration test

hasan7n:
TODO:
enforce user input to use registered uid. Otherwise errors will happen

hasan7n:
A concern to check:
running `medperf run` with a cached experiment and an already submitted result might cause an error

hasan7n:
TODO:
1. fix Execution unittest to check if mlcube.run is called properly (i.e. mlcube params...)
2. Is it worth it to modify compatibility test unittests given we will refactor this command soon?

aristizabal95:
Regarding your second point "modifying unit tests", I'd say do the minimal changes for current tests to be valid. Only add if you feel it's easy to do so. If not, we can start raising our testing priority.

aristizabal95 (approved):
Hey Hasan, it's looking great! I left a few small comments, but otherwise I think is basically done! Thanks for this!

cli/cli_tests.sh
```diff
@@ -289,12 +314,34 @@ echo "\n"
...
 ##########################################################
 
+echo "\n"
+
+##########################################################
+echo "====================================="
+echo "Running outstanding models"
+echo "====================================="
+medperf benchmark run -b $BMK_UID -d $DSET_A_UID
+checkFailed "Model2 run failed"
```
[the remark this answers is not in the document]
hasan7n: you are right! this is a result of copy-pasting earlier blocks :)

PR #323, opened 2022-11-15
Use Medperf Exceptions instead of pretty_error

### LINKED ISSUE #319: [FEATURE] Remove pretty_error from commands
Commands should be reusable in a way that we can easily handle undesired scenarios. Currently, all commands use pretty_error and SysExit. 

Using custom exceptions here will allow for better reusability and more sophisticated commands.

hasan7n:
This PR removes all `pretty_error` calls in the medperf code except in the exception handler decorator. This way, code components can be reusable if needed without the need to catch `SystemExit` exceptions.
Additionally:
1. Exceptions inherit from a base exception `MedperfException` which can optionally accept an argument not to do cleanup (similar to what `pretty_error` does).
2. All our code uses `MedperfExceptions`. Thus, the "unexpected errors" msg now can notify us of a probable bug or that we need custom handling of specific errors.

Closes #319


aristizabal95 (changes requested):
Looks great! Just a small comment regarding excpetions being more specific on our tests. Everything else looks good!

hasan7n:
@aristizabal95 It's now ready.

aristizabal95 (approved):
Looks great! Awesome work Hasan!

PR #326, opened 2022-11-17
Homogenize all and get methods

### LINKED ISSUE #321: [FEATURE] Homogenize Entity.all(), Entity.get()
Benchmark entities use `get` and `all` in a different way than other entities. We want to have analogous implementation across entities, to have at the end the common `Interface` object and hence minimize code maintenance in case of bugs.

Refactor `get` and `all` in all entities.
the following will be done:
1. `all` will get all server entities, and falls back to local ones in case of communication errors.
2. `all` will have an option to use only local.
3. `get` will use `all` with `local` option when falling back to get local entities.

aristizabal95:
This PR standardizes how get and call methods are implemented for each entity, so they follow the same general logic. Now `all` tries first to get all entities from the server, and falls back to local entities if not possible or if the user passes `local_only=True`. Additionally, `get` method no longer calls `all` to find the specific entity.

There's an exception to the above description, which is `Result`. Because of how results are stored, it's not easy to find the desired result in `get` without calling `all`. This is something that should be changed, so results are stored and behave the same as other entities. This should be changed, but is a breaking change.

Closes #321.


aristizabal95:
There are a few topics that must be resolved before continuing with this:
1. Standardizing how entities are stored and retrieved locally. At the moment, datasets and results have different strategies from the rest of the entities.
2. Get all remote results. Should we implement this, even though it will fail for every used except admin?
3. Filtered remote results/entities. How should we reach endpoints like `/me/results` or in general `/me/{entity}`?

aristizabal95:
1. use a temporary local id for entities like datasets and results. When submitted, move them to server uid folder.
2. Just to keep things standard, implement the all method for results. The user command will by default call `--mine`.
3. Have flags for `local_only` and `mine_only` which the user can specify at runtime.

hasan7n:
team discussion:
1. the extra write logic (renaming folders) to be moved to command.submit
2. use tmp uids for all entites:

- dataset: generated hash
- result: bmk_prep_cube
- bmk: name
- cube: name

3. Separate PR: refactoring get like how `all` is refactored (_remote_get, _local_get) **(will create an issue)**
4. Separate PR: sharing code across entities (inheritence, mixins) **(will create an issue)**
5. Possibly a separate PR: compatibility test caching will not be valid because of benchmark tmp uid is dependant on the process id. Introducing the notion of tmp uid in this PR will resolve this (**will check after finalizing the PR**)
6. Discuss with Micah regarding the uniqueness of benchmarks (only the name is unique!)
7. Possibly a separate PR (very unlikely): Check if we need to call the benchmark_models_list when executing the benchmark, as this is no longer called in `bmk.all` (**will check after finalizing the PR**)

cli/medperf/commands/result/create.py
```diff
@@ -68,15 +68,10 @@ def prepare(self):
...
-            self.out_path = results_path(
-                self.benchmark_uid, self.model_uid, self.dataset.uid
-            )
-        else:
-            self.out_path = results_path(
-                self.benchmark_uid, self.model_uid, self.dataset.generated_uid
-            )
+        self.dataset = Dataset.get(self.data_uid)
+        self.out_path = results_path(
+            self.benchmark_uid, self.model_uid, self.dataset.uid
```
hasan7n, line 73: I will dig more into this, but for now it seems that we will get results tmp uids like this: "1_2_None" since we are using the dataset server uid

hasan7n:
@aristizabal95 thanks a lot man for this

Let's merge it? Please take a look at me comments and let me know if these are issues to be open or not

hasan7n (approved):
Thanks a lot man!!

PR #327, opened 2022-11-21
Add priority field for benchmark-model associations

### LINKED ISSUE #206: [FEATURE] Model Priority field for benchmark/model associations
Benchmark owners should be able to set the priority of models within a benchmark in the case of any time considerations (e.g. a challenge, models take 2 weeks to run, etc...). This priority should be editable by the benchmark owner.

hasan7n:
This PR adds a field for the benchmark-model association entity, and an endpoint for the benchmark owner to be able to modify the priorities of their associated models.

Closes #206 

**Clarification**

Priorities in the database are represented by float values. This is to avoid having to shift the priorities of all other benchmark-model associations when updating a single priority.


server/benchmarkmodel/serializers.py
```diff
@@ -114,3 +114,60 @@ def update(self, instance, validated_data):
...
+            raise serializers.ValidationError(
+                "The association should be approved before modifying its priority"
+            )
+        if data["priority_rank"] < -1 or data["priority_rank"] == 0:
+            raise serializers.ValidationError(
+                "The priority value should be a positive integer or -1"
+            )
+        return data
+
+    def update(self, instance, validated_data):
```
hasan7n, line 136: @johnugeorge Do you suggest having this logic on the client rather than the server?

server/benchmarkmodel/serializers.py
```diff
@@ -114,3 +114,60 @@ def update(self, instance, validated_data):
...
+        if min(differences) >= 1e-10:
+            return
+
+        bmk_models = BenchmarkModel.objects.filter(
+            benchmark__id=self.instance.benchmark.id
+        )
+        for i, bmk_model in enumerate(bmk_models):
+            bmk_model.priority = i + 1.0
+
+        BenchmarkModel.objects.bulk_update(bmk_models, ["priority"])
```
hasan7n, line 173: @johnugeorge Do you suggest having this logic on the client rather than the server?

hasan7n: much easier and more REST-friendly to move this logic to the client.

hasan7n:
I will use this PR to set up a scalable testing setting for the server (e.g. seeding the database during `python manage.py test` ...)

cli/medperf/commands/association/priority.py
```diff
@@ -0,0 +1,83 @@
...
+        associated_cubes = [assoc["model_mlcube"] for assoc in self.assocs]
+        if self.mlcube_uid not in associated_cubes:
+            raise InvalidArgumentError(
+                "The given mlcube is not associated with the benchmark"
+            )
+
+        cube_priority = associated_cubes.index(self.mlcube_uid) + 1
+        redundant_command = any(
+            [
+                cube_priority == len(self.assocs) and self.priority >= cube_priority,
```
aristizabal95, line 48: Could you provide some comments to explain these comparisons? I want to also discuss possible edge cases here but I firs need to understand what each check does.

cli/medperf/comms/interface.py
```diff
@@ -40,30 +39,6 @@ def authenticate(self):
...
-
-        Args:
-            benchmark_uid (int): UID of the benchmark
-
-        Returns:
-            Role: the association type between current user and benchmark
-        """
-
-    @abstractmethod
-    def authorized_by_role(self, benchmark_uid: int, role: str) -> bool:
```
aristizabal95, line 55: Any reason why we're removing this? I don't remember where we use it, but I'd like to know the reasoning behind this

aristizabal95 (commented):
Looks good from the client side! just a few minor doubts. I'll leave the server review to Johnu

cli/medperf/comms/interface.py
hasan7n: yeah, there is no reason. Just wanted to remove unused rest methods. lmk if I should revert that back

server/benchmarkmodel/models.py
```diff
@@ -18,6 +18,7 @@ class BenchmarkModel(models.Model):
     approved_at = models.DateTimeField(null=True, blank=True)
     created_at = models.DateTimeField(auto_now_add=True)
     modified_at = models.DateTimeField(auto_now=True)
+    priority = models.FloatField()
```
johnugeorge, line 21: Does Autoincrementing help here? https://docs.djangoproject.com/en/4.0/ref/models/fields/#autofield

hasan7n:
TODO: add integration test

cli/medperf/commands/association/association.py
```diff
@@ -56,3 +57,32 @@ def reject(
...
+    this benchmark is run. A model with a higher priority is executed before
+    a model with lower priority. The order of execution of models of the same priority
+    is arbitrary.
+    Examples:
+    Assume there are three models of IDs (1,2,3), associated with a certain benchmark,
+    all having priority = 0.
+    - By setting the priority of model (2) to the value of 1, the client will make
+    sure that model (2) is executed before models (1,3).
+    - By setting the priority of model (1) to the value of -5, the client will make
+    sure that models (2,3) are executed before model (1).
```
aristizabal95, line 80: Very good explanation! Love this docstring

aristizabal95 (commented):
Looks good from the client side! I'll wait for Johnu to look at the recent changes

server/benchmarkmodel/views.py
```diff
@@ -50,10 +50,15 @@ def get(self, request, pk, format=None):
 
 
 class ModelApproval(GenericAPIView):
-    permission_classes = [IsAdmin | IsBenchmarkOwner | IsMlCubeOwner]
     serializer_class = ModelApprovalSerializer
     queryset = ""
 
+    def get_permissions(self):
```
johnugeorge, line 56: Can we add the permission in the Micah's API doc ?

johnugeorge (approved):

PR #329, opened 2022-11-26
Extend CLI tests

### LINKED ISSUE #285: Extend cli integration tests
It would be better to have `cli.sh` include all possible commands of `medperf`. Some bugs may not be caught using `pytest`.

Would be perfect as a first step to implement the logic of `server/seed.py` inside `cli.sh` but using the command line interface (mainly submitting cubes, submitting a benchmark...)

### LINKED ISSUE #337: [BUG] Compatibility test uses same resullt uid for different experiments
`medperf test -b 1 -e 2` and `medperf test -b 1 -e 3` will produce the same result uid. This is due to the fact that compatibility test is using the benchmark server uid even if the test is modifying the benchmark's component. It should use a temporary benchmark if a benchmark's component is changed.

The solution might be as follows: having a flag component_changed=True/False modified when set_cube_uid changes the benchmak componenets.

hasan7n:
This PR adds more CLI tests to the current `cli/cli.sh` script. Tests include:
1. mlcube and benchmark submission
2. TBD

Closes #285 

**EDIT:** various unexpected bug fixes were encountered due to the new tests. + Fixes #337 

This PR should be ready to be merged. Further tests can be added in other PRs.

 tests to be added:
  * partial results upload
  * priority set (#327 )
  * benchmark.run_all (#318 )
  * lists?


hasan7n:
todo:
1) Discuss with the team if we really need the chexpert integration tests
2) add integration tests for:
    * partial results upload
    * priority set (depends on another PR)
    * benchmark.run_all (depends on another PR)
    * TBD

cli/medperf/entities/cube.py
```diff
@@ -73,24 +73,19 @@ def __init__(self, cube_dict):
...
-
         self.generated_uid = self.name
         path = storage_path(config.cubes_storage)
         if self.uid:
             path = os.path.join(path, str(self.uid))
         else:
             path = os.path.join(path, str(self.generated_uid))
-
+        # NOTE: maybe have these as @property, to have the same entity reusable
+        #       before and after submission
```
aristizabal95, line 83: Here you mean that once submitted, the path no longer reflects what it should be? If that's the case I totally agree! We should make this field dynamic so it gets updated after submission.

cli/medperf/comms/rest.py
```diff
@@ -329,18 +328,18 @@ def get_cube_metadata(self, cube_uid: int) -> dict:
             raise CommunicationRetrievalError("the specified cube doesn't exist")
         return res.json()
 
-    def get_cube(self, url: str, cube_uid: int) -> str:
+    def get_cube(self, url: str, cube_path: str) -> str:
```
aristizabal95, line 331: Was there any specific reason for this change? I like it! It means the comms layer no longer cares how to build the cube path. Did you encounter any issues that got fixed by this? Or was it intuition?

aristizabal95: These changes should be reflected in the comms interface as well. Just copy-paste the signature and docstring

aristizabal95 (approved):
Hey Hasan, this looks great! Loved all the changes you made. Just added a note regarding the changes in the comms methods signature. Just make the small changes on the comms interface and merge this. Great job as always Hasan!

cli/medperf/comms/rest.py
hasan7n: Yes! The reason was this:
In the cube entity, we have download methods. These methods were calling the corresponding comms methods by providing the cube uid. However, after our recent modification of introducing cube uid and cube generated uid, we are now having two possibilities for the path where cube assets are downloaded (before, or,  after submission). We should then in each Cube.download* method check if the cube is submitted then pass the server uid to the comms method, else pass the generated uid.
It seems ugly :) + as you said, we got the benefit of making sure comms should not contain logic of constructing cube path, which is ideal.

hasan7n: This was causing a failure, I forgot where. The cube assets where downloaded in `cubes/<server_uid>` which was actually `cubes/None`

PR #338, opened 2022-12-14
combine list command for all entities

### LINKED ISSUE #320: [FEATURE] One single list command to rule them all
Currently the list command has a separate implementation for each entity. This is something that should be modularized so all entities use the same list logic.

A single common command that receives the specific Entity class, and knows what procedures to call to obtain and print the list of entities.

hasan7n:
This PR refactors the `command.<entitiy>.list` logic by implementing one, general list command that can be used for all four entities.
Closes #320


hasan7n:
**TODO**:
- create association entity (tentative)
- add unit tests

cli/medperf/entities/interface.py
```diff
@@ -52,3 +52,10 @@ def write(self) -> str:
         Returns:
             str: Path to the stored entity
         """
+
+    @property
+    def identifier(self):
+        return self.id or self.generated_uid
+
+    def is_registered(self):
```
aristizabal95, line 60: Loving this! I think this is the kind of cases where implementing the method (or property) inside the abstract class is valid. One thing to think about this is we're indicating we expect `self.id` and `self.generated_uid` to exist in any implementatino of the abstract method. It would be nice to find a way to make this assumption more explicit. (maybe have the interface also inherit from the MedperfSchema later on).

I think we can also have `is_registered` as a property

aristizabal95 (commented):
This looks great! Let me know when the PR is ready to make a final review and approval

cli/medperf/entities/interface.py
hasan7n: Sounds good regarding the schema inheritance! It is better to define the "common" fields explicitly.

And yes! is_registered can be a property. Will do that

hasan7n:
@aristizabal95 thanks for the early review :) I was surprised that I have some local unpushed commits where I modified how we handle `field_mapping`. I think it is cleaner.
Anyway, I will push these changes, do the unit tests and notify you for the final review.

cli/medperf/commands/list.py
```diff
@@ -0,0 +1,54 @@
...
+    def filter(self):
+        self.data = [
+            {field: entity_dict[field] for field in self.fields}
+            for entity_dict in self.data
+        ]
+
+    def display(self):
+        headers = self.fields
+        data_lists = [list(entity_dict.values()) for entity_dict in self.data]
+        tab = tabulate(data_lists, headers=headers)
```
aristizabal95, line 53: Just a note here for future reference. This logic will probably be moved to the UI interface once we establish how to handle arbitrary UI's

cli/medperf/tests/commands/test_list.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # spies
+    generated_entities = [entity_object for _ in display_dicts]
+    all_spy = mocker.patch(
+        "medperf.entities.interface.Entity.all", return_value=generated_entities
+    )
+    ui_spy = mocker.patch.object(ui, "print")
+    tabulate_spy = mocker.spy(list_module, "tabulate")
+
+    system_inputs = {"display_dicts": display_dicts}
```
aristizabal95, line 31: Putting on my "dumb reader" hat on 😆. This variable `system_inputs` isn't clear from a first read. I personally don't know exactly what this means in the context of the tests.

aristizabal95 (changes requested):
Hey Hasan! This looks great. I felt some peace of mind when I saw all the `test_list` files getting deleted haha! I left some comments regarding the tests. 

I'm being a little bit strict with the testing style given that we're trying to establish some common ground for those. I invite you to also be as strict with my tests when it comes to following the new guidelines, as well as challenging me on my comments if you think they don't align, or you think there's a better way to do things. We're going to need a lot of discussion to find a paradigm that's useful, readable and maintainable. Although this looks great, I'll leave those comments as "request changes" mainly to discuss them on our next call

cli/medperf/entities/benchmark.py
```diff
@@ -297,3 +297,19 @@ def upload(self):
         updated_body = config.comms.upload_benchmark(body)
         updated_body["models"] = body["models"]
         return updated_body
+
+    def display_dict(self):
```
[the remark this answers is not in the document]
hasan7n: Sounds great!

cli/medperf/tests/commands/test_list.py
aristizabal95: @hasan7n thinking about this, do you think using `state_variables` would be a more intuitive name? If so it would be nice to change that name in all the tests that use it so far

hasan7n: sounds perfect!

hasan7n:
@aristizabal95 Thanks for the comments! I did the requested changes.

hasan7n:
@aristizabal95 let's merge it :)

aristizabal95 (approved):
Looks great! Merging now!

PR #339, opened 2022-12-14
Test entity interface implementations

### LINKED ISSUE #356: Update Entities unit testing style

aristizabal95:
This PR moves shared tests among Entity implementations into a single file. 

It also implements a different, more manageable solution to tests implementations, where we check the outputs of the functions, instead of the implementation. This means we should start checking for what behavior is expected from a function's output given an input, instead of spying on function calls inside the function of interest.

Closes #356.


hasan7n:
It looks PERFECT!

Suggestions for `test_entities/utils.py`:
1. I think `generate_<entity>` should go to somewhere in the `mocks` folder.
2. About structuring the functions in this file: three suggestions
  a. We can define each entity's setup in its specific test file (e.g. move `setup_benchmark_fs` and `setup_benchmark_comms` to `test_benchmark.py`. `test_entity.py` would then import these setups. The only thing I see is that we have no `test_result.py`. What do you think about creating a `test_result.py` containing only the setups associated with the `result` entity?
  b. split the utils into two files: one for `setup fs` functions and one for `setup comms` functions.
  c. Keep it as it is now :)

A tiny suggestion about the `generate_<entity>` functions: can you use something like the style used below instead of `kwargs.get`?
https://github.com/mlcommons/medperf/blob/fdb4cd222ddf5f033b84684e7b1cedb9cbf23854/cli/medperf/tests/mocks/requests.py#L136-L154

hasan7n (approved):
Thanks, Alejandro!
New year, new and easy testing style :)

aristizabal95:
> It looks PERFECT!
> 
> Suggestions for `test_entities/utils.py`:
> 
> 1. I think `generate_<entity>` should go to somewhere in the `mocks` folder.
> 2. About structuring the functions in this file: three suggestions
>    a. We can define each entity's setup in its specific test file (e.g. move `setup_benchmark_fs` and `setup_benchmark_comms` to `test_benchmark.py`. `test_entity.py` would then import these setups. The only thing I see is that we have no `test_result.py`. What do you think about creating a `test_result.py` containing only the setups associated with the `result` entity?
>    b. split the utils into two files: one for `setup fs` functions and one for `setup comms` functions.
>    c. Keep it as it is now :)
> 
> A tiny suggestion about the `generate_<entity>` functions: can you use something like the style used below instead of `kwargs.get`?
> 
> https://github.com/mlcommons/medperf/blob/fdb4cd222ddf5f033b84684e7b1cedb9cbf23854/cli/medperf/tests/mocks/requests.py#L136-L154

Great suggestions! I'll move the generators to the `mocks` folder. For now I'll keep the setup in the `utils` file as before, since I'm not sure if this setup can be reused outside the `entities` tests. But if we see they can be used outside we could also move them to the mocks folder

PR #342, opened 2022-12-20
Config management open issues

### LINKED ISSUE #334: [BUG] Config options default values show as None on help message
**Describe the bug**
Configuration options that can be passed to every command, or to the profile commands will show default values as `None`, when defaults are specified by the `config.py` file, and usually have assigned default values. This can be misleading/confusing to end-users, as they could believe everything is unset, or not know what are the actual default values without inspecting the code

**To Reproduce**
1. run `medperf --help`
2. look at config options default values

**Expected behavior**
Default values should be displayed to the user.

**Additional context**
This was known at the time of adding the configurable options. The reason this happens is because we need to recognize when an user explicitly passes an option. If using defaults, there's no way to tell which values were passed explicitly and which weren't. Although we could considered default values as unspecified option parameters, this isn't ideal for commands like `profile create` and `profile set`, where the user could want to be explicit about setting a parameter to the currently default one.

hasan7n:
This PR resolves the following:

Closes #334
- Fix how default options appear for users when using `--help`.
- Use custom configuration management instead of `configparser` for better python types handling
- add a `medperf profile delete` command


cli/medperf/config.py
```diff
@@ -53,3 +54,18 @@
 statistics_timeout = None
 infer_timeout = None
 evaluate_timeout = None
+
+configurable_parameters = [
```
aristizabal95, line 58: This is fine, although it would be nice to think of a way of DRYing this up. If we ever decide to change the configurable parameters, we have to do it in two different places, which might be error-prone in the future.

aristizabal95 (approved):
Looks great! I just added a minor detail regarding the UI changes. Feel free to merge once that change is done!

cli/medperf/config.py
hasan7n: Exactly. I will create an issue for this to keep in mind, and add a comment as well.

PR #353, opened 2023-01-05
Explicitly upgrade pyOpenSSL

aristizabal95:
This quickfix updates pyOpenSSL on github action `cli-local` so that it successfully runs the server.


johnugeorge:
Fixed in #354

Issue #352, opened 2022-12-22
[FEATURE] Server and client versioning


hasan7n:
**Is your feature request related to a problem? Please describe.**
Given that some breaking changes have happened (and will keep happening) with the server and the client, a lot of boilerplate code will start showing up just to cover edge cases with backwards compatibility.

**Describe the solution you'd like**

We will follow `{Major}.{Minor}.{Patch}` versioning for both the client and the server:

**Compatibility rules**
- Clients cannot talk to newer servers, e.g. Client 1.1 cannot talk to Server 1.2
- Clients CAN talk to older MINOR version servers, e.g. Client 1.2 can talk to Server 1.1.
- Clients cannot talk to DIFFERENT MAJOR version servers, e.g. Client 2.0 cannot talk to Server 1.13.
- Patch releases are not considered when checking compatibility.
<table class="tg">
<thead>
  <tr>
    <th class="tg-34fe"></th>
    <th class="tg-34fe" colspan="4">Server</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td class="tg-9wq8" rowspan="4">Client</td>
    <td class="tg-c3ow"></td>
    <td class="tg-c3ow">1.0</td>
    <td class="tg-c3ow">1.1</td>
    <td class="tg-c3ow">2.0</td>
  </tr>
  <tr>
    <td class="tg-yj5y">1.0</td>
    <td class="tg-yj5y">Y</td>
    <td class="tg-yj5y">N</td>
    <td class="tg-yj5y">N</td>
  </tr>
  <tr>
    <td class="tg-c3ow">1.1</td>
    <td class="tg-c3ow">Y</td>
    <td class="tg-c3ow">Y</td>
    <td class="tg-c3ow">N</td>
  </tr>
  <tr>
    <td class="tg-yj5y">2.0</td>
    <td class="tg-yj5y">N</td>
    <td class="tg-yj5y">N</td>
    <td class="tg-yj5y">Y</td>
  </tr>
</tbody>
</table>

**How/when versions will be released**

- Major of the server is incremented when there is a breaking change. The client Major version is also incremented to adapt to these changes.
- Minor of the server is incremented when there are backward-compatible changes. Client Minor version is also then incremented to adapt to these changes WHILE preserving compatibility with previous minor releases of the server
- Patch releases for the client are for bug fixes and "soft" changes. Patch releases for the server are also for bug fixes or changes that are not reflected on the client.
[...]

PR #354, opened 2023-01-06
Add versioning support for MedPerf APIs

johnugeorge:
This PR adds versioning support to MedPerf APIs

1. All current URL paths are prefixed with `/api/v1/`
2. `drf_yasg` is replaced with `drf-spectacular` which has support for versioning
3. Added a new api `/version` which returns the server api version

Related: #352 

---
  - https://app.asana.com/0/0/1203888943809394


johnugeorge:
@aristizabal95 @hasan7n Can you make client changes? Server changes are ready

/cc @msheller

server/dataset/tests.py
```diff
@@ -19,7 +19,7 @@ def setUp(self):
         user.save()
         self.client = APIClient()
         response = self.client.post(
-            "/auth-token/", {"username": username, "password": password}, format="json",
+            "/api/v1/auth-token/", {"username": username, "password": password}, format="json",
```
aristizabal95, line 22: I believe it would be best to store the `/api/v1` path on a variable, so that it's easier and less error-prone to edit in the future. Also, it would probably be better to have the current full version be stored on a single place, and use that to build the api path.

This comment applies to all the edits in this file.

server/medperf/urls.py
```diff
@@ -16,38 +16,26 @@
...
-    path("me/", include("utils.urls")),
-    path("auth-token/", obtain_auth_token, name="auth-token"),
+    path("schema/", SpectacularAPIView.as_view(api_version=settings.SERVER_API_VERSION), name="schema"),
+    path("swagger/", SpectacularSwaggerView.as_view(), name="swagger-ui"),
+    re_path(r"^$", SpectacularRedocView.as_view(), name="redoc"),
+    path("admin/", admin.site.urls, name="admin"),
+    path("version", ServerAPIVersion.as_view(), name="get-version"),
+    path('api/', include([
+        path('v1/', include([
+            path("benchmarks/", include("benchmark.urls", namespace="v1"), name="benchmark"),
```
aristizabal95, line 32: Same comment as above but for the namespace. It would be better to store the version somewhere else so that in the future we don't have to look for all the `v1`s.

server/mlcube/tests.py
```diff
@@ -19,23 +19,23 @@ def setUp(self):
         user.save()
         self.client = APIClient()
         response = self.client.post(
-            "/auth-token/", {"username": username, "password": password}, format="json",
+            "/api/v1/auth-token/", {"username": username, "password": password}, format="json",
```
aristizabal95, line 22: Same comment as the previous ones

server/seed.py
```diff
@@ -21,7 +44,7 @@ def request(self, endpoint, method, token, data, out_field=None):
             resp = requests.request(
                 method=method,
                 headers=headers,
-                url=self.host + endpoint,
+                url=self.host + "/api/" + self.version + endpoint,
```
aristizabal95, line 47: This looks close to what I suggest in the previous comments. Something like this, but on the server code would be ideal.

server/user/tests.py
```diff
@@ -14,23 +14,23 @@ def setUp(self):
         password = "admin"
         self.client = APIClient()
         response = self.client.post(
-            "/auth-token/", {"username": username, "password": password}, format="json",
+            "/api/v1/auth-token/", {"username": username, "password": password}, format="json",
```
aristizabal95, line 17: Same comment as previous ones

aristizabal95 (changes requested):
Hey Johnu, thanks for this! It's looking good! I suggest changing the way we're constructing the paths with the version so that it is easier to manage in the future. My suggestion is to have the current full version stored somewhere, and build the api-path prefix using that. This way, we don't need to look for all the paths or namespaces that use `v1` when we want to change the version

aristizabal95:
Hey @hasan7n can you review the changes I made?

hasan7n:
looks good Alejandro! A couple of questions:
1) Can we just use `config.server_url` to contain the version Instead of modifying internal code and appending a version path? I feel it would be better to have it in one place in the config. Not having a solid reason here but I am imagining `medperf profile set --server <some server>` should not still be subject to changing a version prefix... what do you think?
2) In case we didn't go with the above, should we modify the `storage_path` util function to include the server version?

aristizabal95:
@hasan7n I had those doubts as well. 
>  medperf profile set --server <some server> should not still be subject to changing a version prefix
Can you clarify what you mean here? I also feel okay with either solution, but I have some (although weak) arguments for why I decided to go the current route:
- The user can't change the version of the API he's communicating to. If we provide the full path on the server, then he can easily try and connect to an incompatible version. Here instead he needs to go through the code and modify the version variable directly.
- I'd like to use the same url for hosting the server as well as for configuring the client. For example:
```
python server/manage.py runserver 127.0.0.1:8080
medperf profile set --server https://127.0.0.1:8080
```
In the other scenario, the user would need to know how the api path is defined and the version of the server to configure the client, which are both hidden in the code

2. Also thought about this, but if we set the version there then a server changing versions would be reflected as an entirely different server on the client-side, which might or might not be desired. We can discuss this further tomorrow

hasan7n (approved):

PR #366, opened 2023-01-19
Use pydantic models for entities

### LINKED ISSUE #359: [FEATURE] Use pydantic models for entities
Entities have a dictionary definition of their state, and this definition is rewritten multiple times in the code. It would be ideal if there was one source of truth for what an entity expects, which is reused whenever possible

A pydantic model allows a clear and concise definition of the expected state for an object. It additionally provides type and custom validation by default, which may simplify internal code

aristizabal95:
This PR implements pydantic models for medperf entities. This provides validation and explicit definition of expected fields for an entity. Closes #359.


cli/medperf/entities/schemas.py
```diff
@@ -0,0 +1,82 @@
...
+
+class MedperfSchema(BaseModel):
+    id: Optional[Union[str, int]]
+    name: str = Field(..., max_length=20)
+    owner: Optional[int]
+    created_at: Optional[datetime]
+    modified_at: Optional[datetime]
+    # TODO: This must change after allowing edits
+    state: str = "OPERATION"
+    is_valid: bool = True
```
hasan7n, line 16: these should not exist for Result entity

cli/medperf/entities/benchmark.py
```diff
@@ -22,42 +23,38 @@ class Benchmark(Entity):
     what models to run and how to evaluate them.
     """
 
-    def __init__(self, bmk_dict: dict):
+    description: Optional[str] = Field(None, max_length=20)
```
hasan7n, line 26: The difference between how entities behave now and before, is that "forgetting" to pass an optional field wouldn't cause any error. Before this, the constructor would raise key errors if any field is missing. I am just concerned about silent errors. Let me know what do you think

If we are worried about how easy/difficult for us to create temp entities for tests, mocks in the tests folder were already facilitating this.

cli/medperf/entities/result.py
```diff
@@ -22,35 +22,25 @@ class Result(Entity):
...
-
-        dset = Dataset.get(self.dataset_uid)
-        self.generated_uid = (
-            f"b{self.benchmark_uid}m{self.model_uid}d{dset.generated_uid}"
-        )
+        super().__init__(*args, **kwargs)
+
+        dset = Dataset.get(self.dataset)
+        self.generated_uid = f"b{self.benchmark}m{self.model}d{dset.generated_uid}"
         path = storage_path(config.results_storage)
```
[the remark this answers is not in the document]
hasan7n: Or we can merge this now and resolve this when merging main into that PR

hasan7n:
@aristizabal95 Looks awesome! I will approve after we make sure point 1 below is not an issue. I left a few comments above for:

1. Extra fields are defined for the `Result` entity. Is that a problem? This will be written locally.
2. Minor stuff (documentation, maybe extra code piece)
3. A concern about defining the fields as optional
4. Questions that can be addressed in new PRs:
  a. The ID fields in entities currently expect `int` or `str`
  b. After this PR, `Benchmark.tmp` seems to become useless

cli/medperf/commands/benchmark/submit.py
```diff
@@ -45,97 +41,38 @@ def run(cls, benchmark_info: dict, force_test: bool = True):
...
-        demo_url = self.demo_url
-        demo_hash = self.demo_hash
+        data_prep = self.bmk.data_preparation_mlcube
+        model = self.bmk.reference_model_mlcube
+        evaluator = self.bmk.data_evaluator_mlcube
+        demo_url = self.bmk.demo_dataset_tarball_url
+        demo_hash = self.bmk.demo_dataset_tarball_hash
         benchmark = Benchmark.tmp(data_prep, model, evaluator, demo_url, demo_hash)
         _, data_uid, _, results = CompatibilityTestExecution.run(
             benchmark.generated_uid, force_test=self.force_test
```
hasan7n, line 78: I am thinking about passing the benchmark directly to the compatibility test rather than creating a tmp one. This can be in another PR though

cli/medperf/entities/schemas.py
aristizabal95: Created a DeployableSchema. check it out!

hasan7n (approved):
Awesome!

PR #369, opened 2023-01-25
Implement a command to view/store entities

### LINKED ISSUE #349: [FEATURE] commands to view or store individual entities

aristizabal95:
This PR implements a general command to display the contents of an entity or list of entities. It contains the same filtering options as the list command, and also provides a way to directly store the contents to a file. Closes #349.


cli/medperf/commands/view.py
```diff
@@ -0,0 +1,84 @@
...
+        self.entity_id = entity_id
+        self.entity_class = entity_class
+        self.format = format
+        self.local_only = local_only
+        self.mine_only = mine_only
+        self.output = output
+        self.data = []
+
+    def validate(self):
+        valid_formats = set(["yaml", "json"])
```
hasan7n, line 50: just curious :) why set?

cli/medperf/commands/result/result.py
```diff
@@ -56,3 +59,24 @@ def list(
...
+    entity_id: Optional[int] = typer.Argument(None, help="Result ID"),
+    format: str = typer.Option(
+        "yaml",
+        "-f",
+        "--format",
+        help="Format to display contents. Available formats: [yaml, json]",
+    ),
+    local: bool = typer.Option(False, "--local", help="Display local results"),
+    mine: bool = typer.Option(False, "--mine", help="Display current-user results"),
+    output: str = typer.Option(
```
hasan7n, line 76: Let's add this here in the help message?
"Display local results if result ID is not provided"
"Display current-user results if result ID is not provided"

hasan7n: in all four typer commands

cli/medperf/commands/result/result.py
```diff
@@ -56,3 +59,24 @@ def list(
...
+        None, "--output", "-o", help="Output file to store contents"
```
hasan7n, line 77: also here, I think we should add: if not provided, the output will be displayed
in all four typer commands

cli/medperf/tests/commands/test_view.py
```diff
@@ -0,0 +1,142 @@
...
+    local_ids = request.param.get("local", [])
+    remote_ids = request.param.get("remote", [])
+    user_ids = request.param.get("user", [])
+    all_ids = list(set(local_ids + remote_ids + user_ids))
+
+    local_entities = [generate_entity(id, mocker) for id in local_ids]
+    remote_entities = [generate_entity(id, mocker) for id in remote_ids]
+    user_entities = [generate_entity(id, mocker) for id in user_ids]
+    all_entities = list(set(local_entities + remote_entities + user_entities))
+
```
hasan7n, line 45: Let's talk.
I think what we are doing here is "mimicing how entities behave", which we shouldn't do.
Making sure that entity.all is called with the correct mine_only and local_only arguments should be enough for testing this specfic idea. We don't have to trace that all the way to the ui output. what do you think?
I think the entity.list PR illustrates more what I am saying

hasan7n: I want to mention btw that in the priority PR I did, I also mimicked how the server behaves, which should be changed.

hasan7n:
@aristizabal95 all looks great! just some small documentation issues, and there is an issue with the style of the tests (maybe an opinion). Let's discuss it maybe tomorrow.

cli/medperf/commands/view.py
aristizabal95: Because checking for the existence of an element in a set is O(1), although in this specific scenario we don't really gain much given that the check happens only once 😅

hasan7n (approved):

PR #373, opened 2023-02-02
Stream download files

### LINKED ISSUE #276: [FEATURE] Downloading bulk resource files with less memory usage
https://github.com/mlcommons/medperf/blob/c90e7e96dbf7485f7668593f2810f4d36fd447d2/cli/medperf/comms/rest.py#L346-L347
https://github.com/mlcommons/medperf/blob/c90e7e96dbf7485f7668593f2810f4d36fd447d2/cli/medperf/comms/rest.py#L360

Currently, downloading resource files for entities occurs in RAM. Large files may cause problems. An example would be:
A machine learning model that requires multiple model weights files that are not used simultaneously but in multiple inference stages. An example would be some Organ 3D segmentation models.

@aristizabal95 suggestion: We can use this [code](https://stackoverflow.com/questions/16694907/download-large-file-in-python-with-requests) to make file retrieval use streaming procedures to restrict memory usage while downloading files.

hasan7n:
This PR modifies how direct links are downloaded with `requests.get`. Since these links may contain big data, it's better to stream the downloaded data into the disk rather than loading it to the RAM before writing to the disk.

Closes #276

### DIFF
--- a/cli/medperf/comms/rest.py
+++ b/cli/medperf/comms/rest.py
@@ -21,7 +21,7 @@ from medperf.exceptions import (
 
 
 def log_response_error(res, warn=False):
-    # note: status 403 might be also returned if a requested resource doesn't exist
+    # NOTE: status 403 might be also returned if a requested resource doesn't exist
     if warn:
         logging_method = logging.warning
     else:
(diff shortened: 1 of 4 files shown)


aristizabal95 (approved):
Looks great! it was way easier than I thought! Awesome work Hasan 😁

PR #374, opened 2023-02-02
Store Singularity images under shared storage

### LINKED ISSUE #335: [FEATURE] Shared singularity storage folder
For singularity MLCubes, we download the image directly into the `workspace/.image` folder, as per MLCube examples. This works fine, but for situations in which MLCubes share the same exact image, we could be redownloading images multiple times

Having a shared path for all `.sif` images would allow for image reusability. Before downloading an image, we would check if that image already exists in the shared storage and if so we can skip the image download procedure. Images would then be associated to the respective MLCube by creating a symbolic link to the `workspace/.image` path.

**Additional context**
We need to be aware of possible name collisions. The images use the original name provided by the mlcube author, and therefore two cubes with different images but same would overwrite each other.

aristizabal95:
This PR modifies cube image retrieval, so they get stored under a shared `.medperf/.images` folder, identified by their hash. Images are then linked symbolically with the location that mlcube expects to find each image. Closes #335.


hasan7n:
@aristizabal95 Looks great! Everything is fine, but just before approval, I would like to use this PR to have integration tests for running a cube using singularity, given that we will be using singularity very soon and we should make sure nothing breaks.

A couple of observations though. Might be worth tracking in the refactoring issue I just created, if you agree:
1.  I just noticed that the image name in `comms` is inferred from the URL like this: `image_name = url.split("/")[-1]`. I am not sure this is perfect :) what if the URL contained something like `?download=true` or something? do you think this might cause problems?
2. The logic added in this PR sounds to be something that is not a `comms` speciality, right? I also just realized that downloading all the cube files has nothing to do with the server URL, auth token; it is not tied to the REST comms. Maybe we can move such logic to some other place in the future?

hasan7n:
@aristizabal95 Please merge if you find no issues with the integration tests.
I defined the cubes to be compatible both with docker and singularity and added `--platform singularity` to one of the integration tests.
Also fyi we will expect a delay of around 4 minutes for the PR checks. installing singularity takes around 4 minutes.

hasan7n (approved):

PR #378, opened 2023-02-08
File registries / download sources

### LINKED ISSUE #377: [FEATURE] File registries
File download procedure is currently being handled by the communications layer. This doesn't seem to fit with the concern of this layer. A communication layer is in charge of establishing communication between the client and the server.

Have a new interface in charge of downloading files. Implementations of this interface can handle different file sources (Http, XNAT, Synapse) and source authentication.

hasan7n:
Abstraction of the layer that is responsible for downloading assets that are not hosted on the MedPerf server.
This PR should:
- [x]  Provide the ground to add support for multiple file download sources which might need external authentication
- [x]  (probably) Move all mlcube-assets-download logic from `comms` to somewhere else (fixes #377 )

This PR also fix two small bugs:
  - `preds_path` should use dataset generated uid, not its server uid (which might be None)
  - integration tests were using a relative path for local certificate

To Discuss:
- **OPENS:**
  - Decide whether to cache access tokens or username+password.
  - Decide if we want to allow login using only an access token.

---
  - https://app.asana.com/0/0/1203888943809409


hasan7n:
**TODO:**
- [ ] Decide whether to cache access token or username+password. Or allow to provide directly an access token?
- [ ] handle possible auth errors in SynapseLogin class
- [ ] add validators for cube assets links?
- [ ] fix/add unit tests
- [ ] sanity check: test using an additional files tarball hosted on synapse.

hasan7n:
TODO: Documentation?

cli/medperf/comms/entity_resources/sources/synapse.py
```diff
@@ -0,0 +1,58 @@
...
+    @classmethod
+    def validate_resource(cls, value):
+        prefix = SynapseSource.prefix
+        if not value.startswith(prefix):
+            return
+
+        prefix_len = len(prefix)
+        value = value[prefix_len:]
+
+        if re.match(r"syn\d+$", value):
```
msheller, line 25: What if no match? Is this an error that needs raised?

cli/medperf/comms/entity_resources/sources/synapse.py
```diff
@@ -0,0 +1,58 @@
...
+            return value
+
+    def __init__(self):
+        self.client = synapseclient.Synapse()
+
+    def authenticate(self):
+        try:
+            self.client.login(silent=True)
+        except SynapseError as e:
+            msg = "There was an attempt to download resources from the Synapse "
```
msheller, line 35: Is this actually the only possible type of error?

cli/medperf/comms/entity_resources/sources/synapse.py
```diff
@@ -0,0 +1,58 @@
...
+            resource_file = self.client.get(
+                resource_identifier, downloadLocation=download_location
+            )
+        except SynapseError as e:
+            raise CommunicationRetrievalError(str(e))
+
+        resource_path = os.path.join(download_location, resource_file.name)
+        # synapseclient may only throw a warning in some cases
+        # (e.g. read permissions but no download permissions)
+        if not os.path.exists(resource_path):
```
msheller, line 54: Seems a worthwhile note to pass back to our Synapse friends.

cli/medperf/comms/entity_resources/sources/synapse.py
hasan7n: The `validate_resource` methods either return the resource identifier or `None` in case of no match. The function `__parse_resource` in `utils.py` will raise an error if it doesn't receive any resource identifier when calling `validate_resource` from all supported sources

cli/medperf/config.py
```diff
@@ -1,11 +1,12 @@
 from os.path import expanduser, abspath
+from pathlib import Path
 
 version = "0.0.0"
 server = "https://api.medperf.org"
 certificate = None
 
 local_server = "https://localhost:8000"
-local_certificate = "server/cert.crt"
+local_certificate = str(Path(__file__).parent.parent.parent / "server/cert.crt")
```
msheller, line 9: Why did this change have to be made? I'm worried that the fix might also fail for the same reason at some point. These kinds of bugs can be so annoying :)

cli/medperf/comms/entity_resources/sources/synapse.py
hasan7n: Perhaps I should document the new code

cli/medperf/comms/entity_resources/sources/synapse.py
hasan7n: All exceptions from the synapse client inherit from `SynapseError`.
I am following a convention that we only catch errors we know why they happen

cli/medperf/config.py
hasan7n: I see :)
I was looking again and I think this change is redundant. The error I faced seems to originate from another place.

hasan7n:
TODO: rename `Public` to `DirectLink`. More precise and correct naming

msheller (approved):
Looks good. I had a few minor questions, but I don't think they need to hold up the merge. Thanks, Hasan!!

cli/medperf/comms/entity_resources/sources/synapse.py
msheller: Sorry, I wasn't clear. I meant is this the only reason we would get a SynapseError, or could there be other reasons for a SynapseError?

cli/medperf/tests/commands/auth/test_synapse_login.py
```diff
@@ -0,0 +1,79 @@
...
+    # Arrange
+    mocker.patch.object(ui, "prompt", return_value=user_input)
+    token_spy = mocker.patch(PATCH_LOGIN.format("SynapseLogin.login_with_token"))
+    pwd_spy = mocker.patch(PATCH_LOGIN.format("SynapseLogin.login_with_password"))
+
+    # Act
+    SynapseLogin.run()
+
+    # Assert
+    if user_input == "1":
```
aristizabal95, line 52: Although branching is usually discouraged, I think in this scenario is fine. We should document what `"1"` and `"2"` mean in this case

cli/medperf/tests/comms/entity_resources/sources/test_synapse.py
```diff
@@ -0,0 +1,103 @@
...
+    mock = mocker.create_autospec(synapseclient.Synapse)
+    mocker.patch.object(synapseclient, "Synapse", return_value=mock)
+
+
+def test_download_works_as_expected(mocker, fs, synapse_client):
+    # Arrange
+    outpath = "path/to/out.tar.gz"
+    contents = "contents"
+    syn_id = "syn535353"
+    synapse_source = SynapseSource()
```
aristizabal95, line 28: This arrange block above seems to be repeated on multiple tests. It might be best to put it inside a setup fixture for those tests

aristizabal95 (commented):
Hey Hasan, this is looking good. I'm mainly concerned with some tests that seem a little bit brittle and hard to read. Let me know if you agree and we can have a discussion about this next week. I want to find a good balance between test robustness and required effort

cli/medperf/comms/entity_resources/sources/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .direct import DirectLinkSource
+from .synapse import SynapseSource
+
+supported_sources = [DirectLinkSource, SynapseSource]
+
+__all__ = [supported_sources]
```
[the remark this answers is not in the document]
hasan7n: [Check this](https://stackoverflow.com/questions/44834/what-does-all-mean-in-python)
I changed how our `__all__`'s are defined. It seems we don't actually get the benefit of it since we don't use star imports, but it is useful to suppress flake's errors of `imported but unused`.

PR #391, opened 2023-02-24
Add requests logging

aristizabal95:
This PR adds requests logging to our logging function. It also separates the setup logging procedure into its own function for readability.


hasan7n (approved):
Looks good!

PR #396, opened 2023-03-07
Implement Cookiecutter templates for relevant MLCubes

### LINKED ISSUE #392: [FEATURE] Have cookiecutter templates for medperf-related MLcubes
Creating a new MLCube requires a lot of steps, and for medperf we have many specifications that are time-consuming for a user to specify correctly. 

We can provide empty templates for creating MLCubes using cookiecutter, so they can create a medperf-compatible MLcube with just one command

aristizabal95:
This PR implements Cookiecutters for own MLCube specifications. It specifically provides templates for the Data Preparator, Model and Evaluator MLCubes. Additionally, it adds a command to the medperf CLI to implement those templates through

```
medperf mlcube create <TEMPLATE_NAME>
```

Closes #392.


cli/medperf/config.py
```diff
@@ -70,3 +70,10 @@
     "platform",
     "cleanup",
 ]
+
+github_repository = "https://github.com/aristizabal95/medperf-2"
+templates = {
```
hasan7n, line 75: Is there a reason for using a repo? (which needs internet connection)
we can specify the folder: `abspath(Path(__file__).parent.parent.parent)`

cli/medperf/commands/mlcube/mlcube.py
```diff
@@ -28,6 +29,27 @@ def list(
     )
 
 
+@app.command("create")
+@clean_except
+def create(
+    template: str = typer.Argument(
+        None,
+        help=f"MLCube template name. Available templates: [{' | '.join(config.templates.keys())}]",
```
hasan7n, line 37: this should be a required argument, right?

hasan7n (changes requested):
Looks great! I left two small comments

cli/medperf/config.py
aristizabal95: There is actually! We expect people in the future will install medperf with `pip install medperf` instead of cloning the repository. In those cases, the templates would not be copied, so providing the repository covers those cases.

aristizabal95: Maybe we can add the templates to the package folders, and make the change you suggested. I'll see if that's possible

cli/medperf/commands/mlcube/mlcube.py
aristizabal95: You're right! I'll make the required change

cli/medperf/config.py
aristizabal95: That absolutely worked, great idea Hasan!

PR #400, opened 2023-03-22
Compatibility tests refactoring and opens

### LINKED ISSUE #371: Move CompatibilityTest retrieval and storage logic to a "TestReport" entity
At the moment, we have special logic for when a benchmark execution is done with not submitted elements. We want to move this into a new entity called TestReport. This entity behaves almost the same as all entities, except that it can't interact with the server. This requres some changes to the Unit test logic.

hasan7n:
This PR addresses the following regarding the compatibility test command:

- Split the logic in `compatibility_test.py` into:
  - [x] `entities.test_report` which contains the logic of interfacing with the file system
  - [x]  `compatibility_test.validate_params` which contains logic for validating inputs params for this particular command
  - [x]  `compatibility_test.run` which contains the main command code (preparing and running the test)
  - [x] `compatibility_test.utils` which contains stateless logic that doesn't need to be in the main `compatibility_test.run` file

It also adds the following features:
  - [x]  Allow directly passing demo dataset url to the command, which removes the necessity of providing a benchmark in all the scenarios of using the command
  - [x] Allow testing from raw data (unprepared data)
  - [x] Allow passing a `--offline` options to instruct the client to rely only on the local storage for testing (removes the necessity of internet connection). To do this, `entity.get` commands now have a `local_only` option.
  - [x]  Allow listing and viewing test reports

Additionally:
  - it fixes a bug preventing testing with local mlcubes.
  - It allows `data_preparation_mlcube` field in `Dataset` locally to accept a string. This was necessary to allow testing with local mlcubes (as this was the first scenario we prepare a dataset with an unregistered mlcube)

Closes #371


hasan7n:
TODO:
- [ ] code documentation
- [ ] unit tests

hasan7n:
use temp uid of the cubes as hash of the contents

hasan7n:
inheritence vs. the validator I made: ask Micah

hasan7n:
unit tests for `TestReport` to be postponed after we have common implementations at the entity interface.

cli/medperf/commands/compatibility_test/run.py
```diff
@@ -0,0 +1,243 @@
...
+            evaluator,
+            data_path,
+            labels_path,
+            demo_dataset_url,
+            demo_dataset_hash,
+            data_uid,
+            no_cache,
+            offline,
+        )
+        test_exec.validate()
```
aristizabal95, line 85: This is one of the things that caused confusion to me. I couldn't find the `validate` implementation for a while, until I realized it was being inherited. Inspecting the logic of this class requires inspecting the logic of another class to fully understand this behavior

cli/medperf/commands/compatibility_test/run.py
```diff
@@ -0,0 +1,243 @@
...
+        self.labels_path = labels_path
+        self.demo_dataset_url = demo_dataset_url
+        self.demo_dataset_hash = demo_dataset_hash
+        self.data_uid = data_uid
+        self.no_cache = no_cache
+        self.offline = offline
+
+        # This property will be set by the validator class
+        # to either "path", "demo", "prepared", or "benchmark"
+        self.data_source = None
```
aristizabal95, line 124: This was also confusing. It is not known from reading this file at which time and how this property is being assigned. It would be better to have this call something like `self.data_source = self.validator.get_data_source()`

cli/medperf/commands/compatibility_test/validate_params.py
```diff
@@ -0,0 +1,106 @@
+from medperf.exceptions import InvalidArgumentError
+
+
+class CompatibilityTestParamsValidator:
+    """Validates the input parameters to the CompatibilityTestExecution class"""
+
+    def __validate_cubes(self):
+        if not self.model and not self.benchmark_uid:
```
aristizabal95, line 8: This is another example of what I meant by highly coupled implementation. This class by itself doesn't have any definition of the properties being tested, so reading this class by itself makes it appear like it will always raise exceptions

cli/medperf/commands/compatibility_test/validate_params.py
```diff
@@ -0,0 +1,106 @@
...
+            )
+        if any([self.data_path, self.labels_path, self.data_uid]):
+            self.__raise_redundant_data_source()
+
+        if not self.data_prep and not self.benchmark_uid:
+            raise InvalidArgumentError(
+                "A data preparation cube should be passed when specifying raw data input"
+            )
+
+    def __validate_data_source(self):
```
aristizabal95, line 64: This function appears to be doing two separate things: validating the parameter combination, and assigning the data source. I would separate this into two functions, and inside the compatibility test class, call `self.data_source = self.validator.get_data_source()`

aristizabal95 (changes requested):
Awesome work Hasan! Lot's of QOL improvements in this repo, thanks for that! I only have a suggestion on how the compatibility test validation is being done. I personally think it would be better to strive for composition instead of inheritance in this occasion. It feels like inheritance is not being used in its intended purpose conceptually, and it can get confusing and hard to read for other developers.

BTW I loved how the tests are looking! Most of them are short, clear and to the point! Thanks for that effort there!

cli/medperf/entities/report.py
```diff
@@ -0,0 +1,138 @@
...
+import logging
+from typing import List, Union, Optional
+
+from medperf.utils import storage_path
+from medperf.entities.schemas import MedperfBaseSchema
+import medperf.config as config
+from medperf.exceptions import InvalidArgumentError
+
+
+class TestReport(MedperfBaseSchema):
```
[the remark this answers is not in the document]
hasan7n: I see. I strongly believe that there is no need for branching for the entity classes like we have for the schemas, especially because we have only 5 implementations, not that much. I wonder how neat that will appear. If we want to inherit from Entity here I would suggest then simply not implementing the unsupported methods, or implement them with raising some nonimplementederror or something similar.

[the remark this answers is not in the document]
hasan7n: I think I was wrong. It seems best practice not to overwrite the upload method as I suggested, and a mixin will be the only option :)

aristizabal95 (approved):
Looks beautiful! Thanks for considering the suggestions and applying them. Feel free to merge it!

PR #404, opened 2023-03-22
List benchmark-specific results

### LINKED ISSUE #397: [FEATURE] Provide command to list all results related to a benchmark
We can list all entities of a certain type, but we can't filter these lists. Additionally, for some lists, there are permissions that block users from retrieving the whole list, like the `Result` entity list. This can be really problematic for benchmark owners who want to retrieve results for a specific benchmark, since currently they don't have the option to do so. 

We need to provide a command that uses the `benchmarks/{id}/results` endpoint to retrieve the list of results.

aristizabal95:
This PR implements the required changes to allow listing results from a specific Benchmark. This was done by modifying the `all` logic to use composition. That way, we can pass specific communication functions if needed (as in this case) while retaining the current default behavior. Closes #397.


cli/medperf/utils.py
```diff
@@ -62,6 +62,23 @@ def read_credentials():
...
+    write_config(config_p)
+
+
+def get_current_user():
+    config_p = read_config()
+    try:
+        current_user = config_p.active_profile["current_user"]
+    except KeyError:
+        raise ExecutionError(
+            "Couldn't retrieve current user information. Please login again"
```
hasan7n, line 77: Maybe a generic MedperfException instead of ExecutionError?

hasan7n (changes requested):
Looks great! In addition to the comment I left,

the filters seem to work only for remote calls. Users running `result ls --benchmark 1` will be surprised to see unfiltered results if the connection was lost. I can see that this is a rare case, but we should at least capture this somewhere. Maybe mention this in the command's description if we don't want to address it in this PR.

aristizabal95:
@hasan7n Totally! That was intended to keep PR concerns separate. We can create a new issue to implement local filtering.

hasan7n (approved):

PR #409, opened 2023-03-30
Filter sensitive information in logging

### LINKED ISSUE #406: [BUG] Debug logging shows login password
**Describe the bug**
Since piping `requests` logging, the body of a POST request is added to the logs. This is also the case for the `auth-token` endpoint, which receives username and password.

**To Reproduce**
1. Login to medperf using the `--loglevel=debug` option
2. Check logs

**Expected behavior**
Sensitive endpoints should either contain filters for masking the password, or not be logged at all

aristizabal95:
This PR implements the basic logic to filter out anything that looks like a password or token from the logs. Closes #406.


hasan7n (approved):
Looks great!

PR #417, opened 2023-04-19
Windows compatibility

### LINKED ISSUE #363: Windows Support

aristizabal95:
CThis PR applies the necessary changes to provide support for Windows machines. It does so by adjusting requirements, updating the MLCube packages, and providing additional logic if necessary. Closes #363.


cli/medperf/entities/cube.py
```diff
@@ -295,7 +300,7 @@ def run(
         proc_out = combine_proc_sp_text(proc)
         proc.close()
         logging.debug(proc_out)
-        if proc.exitstatus != 0:
+        if proc.exitstatus and proc.exitstatus != 0:
```
hasan7n, line 303:
```suggestion
        if proc.exitstatus:
```
:)

cli/requirements.txt
```diff
@@ -1,18 +1,20 @@
-typer~=0.6.0
+typer~=0.6.0; sys_platform != "win32"
+typer<0.5.0; sys_platform == "win32"
 rich~=12.5.0
 PyYAML>=5.4,<6
 requests>=2.26.0
 pydantic==1.10.2
 yaspin==2.1.0
 tabulate==0.8.9
 pexpect==4.8.0
```
hasan7n, line 9: do we need this for windows?

hasan7n (changes requested):
Great work Alejandro, thanks! A couple of points:
1. Let's add Windows OS for the workflow runner configuration. Do we need to contact administrators about this before?
2. Since singularity is not supported in Windows, do we have an informative message when someone tries to use `--platform singularity` on Windows?
3. I am a bit concerned about using `wexpect`. `pexpect` mentions it but says that it is **unmaintained**. I wonder if we can later see an alternative (even maybe an alternative to `pexpect`?)

cli/requirements.txt
hasan7n: here I meant to refer to `pexpect`

hasan7n:
the only todo: workflow for windows

hasan7n:
closing as outdated

PR #420, opened 2023-04-21
Provide a parameter for specifying GPUs to use

### LINKED ISSUE #238: [FEATURE] Option for easily assigning a GPU to an execution flow when multiple GPUs are present
Currently, medperf doesn't handle anything GPU related, leveraging this to MLCube. This has caused some issues during the FeTS Challenge, where some contributors want to be specific about which GPU to use. 

There should be an option that can be passed to MLCube to specify which GPU to use. Something like `--gpu=0` which would be piped to `Cube.run` and handled by MLCube. We must first determine what MLCube provides for this kind of scenarios.

aristizabal95:
This PR allows users to specify what GPUs to expose to a medperf execution. Closes #238.


hasan7n:
if someone specified --gpus but the accelerator count was 0, what do you think about we overriding the accelerator count?

hasan7n:
Everything looks great! But I think we should add some description about how the parameter works right? to cover the following: what happens when the param is None, how to specify the gpus? and also we should note that specifying gpus is ignored if the mlcube to be run was configured with zero accelerator count (this is what happens right?)

aristizabal95:
Hey @hasan7n, I added more details about this parameter on the param help. Let me know if that would be enough! I wasn't sure if it would be worth it to write a whole CLI website on this PR only to cover the GPU parameter details.

Issue #414, opened 2023-04-19
Permission denied when medperf tries to delete folders generated from docker mlcubes


hasan7n:
According to how docker is run by default, the root user owns folders generated by docker MLCubes, and hence medperf can't delete such folders.

MedPerf MLCubes generate root-owned folders only in these two steps:
- `infer` task, if the MLCube generates predictions folders inside the main output mount point (`predictions`).
- `prepare` task, if the MLCube generated prepared data folders inside the main output mount point (`data_path`).

Note: the problem doesn't happen when:
- the outputs are only files
- if the MLCube uses the singularity runner

**Why MedPerf needs to delete such folders? Two possible scenarios:**
- cleanup after a failed inference experiment. For example, if the metrics cube failed, medperf will try to cleanup the predictions folder.
- Later when we add new commands for users to delete entities (deleting a prepared dataset).

**Possible solutions?**
- We may for now catch this permission error and instruct the user to manually delete these files. We can't totally skip cleanups; clutter paths might cause problems (e.g. when listing local entities, or reruning, ...)
- There is a possibility to instruct the MLCube to run the docker image using the current user (tested):
`-Pdocker.cpu_rgs = $(id -u ${USER}):$(id -g ${USER})`
but are there any complications? (possibility of losing permissions inside the docker container, security?...)

aristizabal95:
It is possible to pass information to the docker runner so it runs on behalf of an user (second solution you provided). The MLCube team will see if it's okay to have that by default when running Docker MLCubes, instead of defaulting to the root user

hasan7n:
Since medperf will have sufficient permissions to rename such objects (#422 ), let's not use the solution of running docker container with the current user id as we don't know what implications this may do (e.g. user will have less permissions)

hasan7n:
closed by #422

Issue #365, opened 2023-01-19
[FEATURE] Better cleanup


hasan7n:
**Is your feature request related to a problem? Please describe.**
1. Prepared demo datasets are supposed to be deleted. Those generated after mlcube associations are not deleted.
2. Empty `tmp{pid}` are constantly being created in medperf storage. Cleanup fails to reach all of them due to the {pid} uniqueness.

Solution: we may consider calling cleanup at the end of the process execution, even if the process succeeds.

hasan7n:
++
Cleanup of specific paths should be disabled if the user passed `--no-cleanup`. Example:
https://github.com/mlcommons/medperf/blob/73b5ffc209273987bf68648c06278b2a1d309f63/cli/medperf/commands/execution.py#L75

This was the case until we started to use `cleanup_path`. The check `if config.cleanup` should be put also in `cleanup_path`

hasan7n:
closed by https://github.com/mlcommons/medperf/pull/422

PR #422, opened 2023-04-24
Remove tmp prefixes and better cleanup mechanism

### LINKED ISSUE #408: [FEATURE] Remove TMP prefix usage
Temporary prefixes are being used to identify an entity that we don't want to submit to the server accidentally. This temporary uid has some concurrency issues which are being solved in a hacky way. We can instead identify this entities with a local field that says if they're intended to be submitted or not.

hasan7n:
This PR will change how cleanup is handled:

1. It removes all temporary prefixes that were used to identify what files should be cleaned up after a process is finished. Instead:
    - All temporary folders/files should be created in `config.tmp_storage` if applicable.
    - If not applicable (e.g. temporary MLCubes that need to be present in the MLCube storage), a global list (`config.tmp_paths`) will be used to keep track of such objects
2. Prepared demo datasets will no longer be deleted, and `Entity.upload` now raises an error when someone attempts to upload test entities such as demo datasets.
3. All code attempts to delete objects from the file system, including cleanup, will be handled by `utils.remove_path`. This function will fall back to move the target object to a `.trash` folder in case of deletion failure (e.g. due to permission error). We assume that medperf will always have rename permissions since all objects medperf will attempt to rename live in folders created by itself.
4. Cleanup will be called at the end of each process anyway.

**NOTE**
Cleanup of failed `cube.get` should be handled by `resources.download`, which will be addressed in a separate PR.

Closes #408 #365 #414


aristizabal95 (approved):
Looking good! I'm glad to see a lot of code for handling errors and cleanups has been removed. This is great Hasan thanks!

PR #425, opened 2023-04-27
Do logging rollover per command execution

aristizabal95:
This simple PR forces logging files to be created per execution, instead of by file size. This could make it easier for people to share the logs that are important for events or errors they encountered.

I was planning on changing the name of the logs so they reflect the subcommands executed, but ti wasn't possible to fully extract the subcommands without having to implement a full argument parsing solution, and also removed the benefits of handling the maximum number of log files. Instead, the command executed is being logged at the beginning of every logfile


hasan7n (approved):
Nice!

PR #426, opened 2023-05-02
Use the correct endpoint for user mlcubes

hasan7n:
This PR fixes a small bug where `Cube.remote_prefilter` was using user results endpoint instead of user cubes endpoint


aristizabal95 (approved):

aristizabal95:
Nice catch!

PR #428, opened 2023-05-03
Use Singularity compatible reference examples

### LINKED ISSUE #415: Singularity-compatible MLCube templates and examples
Singularity differs from Docker regarding the notion of `WORKDIR`. Singularity images, when being converted from docker images, ignore the `WORKDIR` instruction. The current user's home directory will be the default working directory. This causes a problem if the Dockerfile is configured using an entry point that uses relative paths.

**What should we do?**

It is better to have Dockerfiles of our templates and examples configured such that the entry point uses an absolute path. For example, don't use this:

https://github.com/mlcommons/medperf/blob/96377eeaf90e5a87f98309e125dc6b4a0b146338/cli/medperf/templates/model_mlcube/%7B%7Bcookiecutter.project_slug%7D%7D/project/Dockerfile#L29-L31

But instead: `ENTRYPOINT ["python3", "/project/mlcube.py"]`

**Also, make a note in the documentation.**

hasan7n:
This PR makes sure ChestXray example and MLCube templates are compatible with Singularity. Closes #415


aristizabal95 (approved):
Looks great!

PR #435, opened 2023-05-16
Format and print dictionary error strings

### LINKED ISSUE #274: Better Logging in case of duplicate Dataset submission
Currently submitting the same dataset is not allowed. The reason we have this is that we want to avoid dataset collision as much as possible. A few edge cases, however, might happen while using MedPerf, and the user will not be aware that they are submitting a duplicate dataset:

1. **Datasets of the same input and output hashes:** Having different users testing the platform and using some public datasets.
2. **Datasets of different input hashes and same output hashes:** A user preparing their dataset with a different labels format and different prep cube, and unaware that the same dataset will be generated.

We should be more verbose and describe better what happened, why that is not allowed, and describe possible reasons why the user might end up with the same generated prepared data.

aristizabal95:
This PR provides more verbose error messages for entity creation errors and REST errors, by formatting the returned error message (dictionary of field-errors) into a human-redeable string. Closes #274.


hasan7n (changes requested):
Nice UI enhancement!

The logic of parsing the error dict however doesn't seem to work always. I tried an unauthorized request to submit a result and got the following:

```
❌ Could not upload the results: 
- detail: 
        - Y
        - o
        - u
        -  
        - d
        - o
        -  
        - n
        - o
        - t
        -  
        - h
        - a
        - v
        - e
        -  
        - p
        - e
        - r
        - m
        - i
        - s
        - s
        - i
        - o
        - n
        -  
        - t
        - o
        -  
        - p
        - e
        - r
        - f
        - o
        - r
        - m
        -  
        - t
        - h
        - i
        - s
        -  
        - a
        - c
        - t
        - i
        - o
        - n
   `

aristizabal95:
@hasan7n This should be fixed now. I also added tests for the different outputs I've encountered. Let me know if it looks good now!

hasan7n (approved):
Looks good!

PR #439, opened 2023-05-18
Make install from subdirectory explicit in installation docs

AlexanderGetka-cbica:
Changes installation.md in docs to make the pip install step explicitly point to the subdirectory. 
When installing on my machine, the step as written looked for a PyPI package called "cli" -- thankfully none exists, but if it did, it would appear as a success to the user.


hasan7n (approved):
Thanks @AlexanderGetka-cbica !

PR #451, opened 2023-06-10
Tutorial doc updates

vpchung:
## Changelog
* ~specify an older version of setuptools (66.1.0) so that importing synapseclient will work~ ← EDIT: noticed that @hasan7n also added a fix for this issue; I will remove my change to prevent merging conflicts
* reduce `submit`'s description  value to <20 characters in order to pass validation check:
  ```
  ❌ Field Validation Error:
  - description: ensure this value has at most 20 characters.
  ```
* fix filename in docs for running the dev server (unless underscores are desired?  Then renaming the script will be needed)
* add `curl` option to `tutorials_scripts/setup_benchmark_tutorial.sh` and `tutorials_scripts/setup_data_tutorial.sh` for users who don't have `wget`
* fix (?) broken elements called in the model_owner_demo README:
Screen Shot 2023-06-13 at 11 20 02 PM
   > **Note**: I don't actually know if this fixed it, as I don't see a change in my local preview.

### DIFF
--- a/cli/requirements.txt
+++ b/cli/requirements.txt
@@ -16,5 +16,5 @@ mlcube-singularity==0.0.9
 validators==0.18.2
 merge-args==0.1.4
 synapseclient==2.7.0
-setuptools==66.0.0
+setuptools==66.1.0
 schema==0.7.5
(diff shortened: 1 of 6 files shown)


hasan7n (approved):
Thanks @vpchung !

PR #452, opened 2023-06-13
Auth0

hasan7n:
Implement authentication with [Auth0](https://auth0.com/).

User identity now will be managed by Auth0. Users' emails and passwords are no longer stored in MedPerf's database. Users no longer require passwords. We use [passwordless authentication](https://en.wikipedia.org/wiki/Passwordless_authentication)

## Client Auth Commands

- **Login**: Using the client, users will request access tokens from Auth0 to be able to access the MedPerf server. An access token is retrieved through the [device authorization flow](https://datatracker.ietf.org/doc/html/rfc8628).
- **Logout**: Users will be able to clear their medperf client's login state. This will remove their tokens from the storage and revoke refresh tokens.

## How will the users interact with the device authorization flow? (login)
_(All UI pages mentioned below are Auth0's built-in pages. They can be customized e.g. by putting our logo...)_
- The user will run `medperf auth login`. They will be prompted through the CLI as follows:
```
Please go to the following link to complete your login request:
        https://dev-5xl8y6uuc2hig2ly.us.auth0.com/activate?user_code=TFCX-CRZP

Ensure that you will be presented with the following code:
        TFCX-CRZP
```
- The user can open the link on any device they want.
- When the link is opened, the user will be prompted by an Auth0 page to confirm that the user code they see on the page is the same one as the one presented in the medperf CLI.
- After confirmation, the user will be redirected to another page to authenticate with the Auth0 backend.
- Meanwhile, the medperf client will be polling the access token from an auth0 endpoint. The access token will become available once the user successfully authenticates.

(A more complete description of the flow can be found [here](https://auth0.com/docs/get-started/authentication-and-authorization-flow/device-authorization-flow))

## What happens on the MedPerf server?

**Authentication**
The MedPerf server has nothing to do with all mentioned above.

**Resource requests**

Any request sent to the server must have an authorization header containing an access token. The server will validate and decode the token (using [this library](https://django-rest-framework-simplejwt.readthedocs.io/en/latest/) which is a Django wrapper around the industry standard [PyJWT](https://pyjwt.readthedocs.io/en/stable/)) and extract the token's subject ID (which will act as the user's Auth0 unique identifier). The user database in medperf server will contain a unique field storing this subject ID. Users will be retrieved using this ID, or created if they don't exist. User objects in our server will still have the main database ID that is used to link the user with our benchmark entities.

This is the only way users can access the MedPerf server. For the Django admin user, see the section about this later below.

## Email Verification by Auth0

This is no longer needed since we are using passwordless authentication (and signup)

<del>
After successful signup, Auth0 sends an email verification link. User object field "email_verified" (boolean) will become True when the user verifies.

Auth0 by default keeps letting unverified users sign in and retrieve access tokens. How do we mitigate that?

**Auth0 Custom Action**

An access token will never be issued if the user tries to login before verifying their email address. This is NOT an Auth0 built-in feature. It's a [custom piece of code](https://auth0.com/docs/customize/actions/flows-and-triggers/login-flow#login-post-login) executed after the user authenticates in Auth0 and before an access token is issued.

**Auth0 Authentication Sessions will still be created**

It might be worth to note that an **Auth0 session** will be created anyway when an unverified user logs in, but that shouldn't be a concern. This session will be of no use for the user.

**An edge-case scenario?**

If a malicious user was trying to use an email they don't own, or some user mistakenly used an email they don't own. What should the real email owner do if they want to signup?

- Ideally, they will reach out to us and we will delete the account to let them signup again!
- Resetting their password is not a perfect solution. The reason is that the first user, by their signup request, may have supplied some name, profile picture, or phone number. Such information cannot be changed without us (there is a paragraph about this below), and will stay hanging there. It doesn't hurt since medperf will not use it, but not right.
</del>

## User Data

**In Auth0:**
Auth0 stores the following data for users:
```
{
    "created_at": "2023-05-23T11:04:55.029Z",
    "email": "hasankassim7@hotmail.com",
    "email_verified": true,
    "identities": [
        {
            "connection": "Username-Password-Authentication",
            "provider": "auth0",
            "user_id": "646c9dd6f7fb73d04c8ca159",
            "isSocial": false
        }
    ],
    "name": "hasankassim7@hotmail.com",
    "nickname": "hasankassim7",
    "picture": "https://s.gravatar.com/avatar/2d3d2eff52c5b5842a90139af743f48b?s=480&r=pg&d=https%3A%2F%2Fcdn.auth0.com%2Favatars%2Fha.png",
    "updated_at": "2023-06-11T19:17:04.981Z",
    "user_id": "auth0|646c9dd6f7fb73d04c8ca159",
    "blocked": false,
    "last_password_reset": "2023-06-11T19:16:04.495Z",
    "last_ip": "2a01:cb10:5ac:f800:eda3:5824:1101:ba18",
    "last_login": "2023-06-11T19:17:04.981Z",
    "logins_count": 16,
    "blocked_for": [],
    "guardian_authenticators": []
}
```
On the MedPerf server, nothing will be stored except the `user_id` which will be used to link Django user objects with the Auth0 user object above.

In our current logic, there is no use-case for retrieving or editing the user data.

For possible reasons in the future:

- **Sending emails to users for notifications**: User's email will be present in access tokens. This is NOT an Auth0 built-in feature. It's a [custom piece of code](https://auth0.com/docs/customize/actions/flows-and-triggers/login-flow#login-post-login) executed to insert the email to issued access tokens. Why like this: 
    - It's better not to store the email on the medperf server to avoid having same information stored in multiple database and worry about sync.
    - It's better not to call the `/userinfo` Auth0 endpoint because it seems to be designed for clients (highly rate limited 5 requests per minute per user id)
    - By design and for some security measures, ID tokens (that contain information such as user email and can be retrieved during login) are not meant to be decoded and used in resources servers.
    - The current approach of inserting the email as a custom claim in the access token complies with [OAuth2.0 specifications](https://datatracker.ietf.org/doc/html/rfc9068#name-identity-claims), with the only concern mentioned there is the privacy of users' email addresses being public (which is far beyond our design concern).

- **Web UI**: User data can be retrieved in the web client by asking for an ID token during login beside the access token. It's worth to note though that user personal data like name or profile picture cannot be edited by the users (weird!) but only by Auth0 admins. We might then at that time think if we want this data to be stored on our server and not on Auth0.

## Refresh tokens

- Access tokens will be short-lived. Refresh tokens are returned with access tokens. The medperf client will refresh the access tokens when necessary.
- Refresh tokens will be renewed each time used, and previous one will be invalidated (for better security)
- Refresh tokens won't work when a user is blocked, or deleted.
~~- Refresh tokens ARE NOT revoked when the user reset their password.~~ Not relevant since we started using passwordless
- A refresh token is invalidated if there was an attempt to use an older (hence invalid) refresh token.
  - Invalidation due to refresh token rotation works on one "series" of refresh tokens not on all refresh tokens of the user. When the user logs in again ( or one some other device), there will become two refresh tokens active. They are independent in the sense that if the first refresh token was used then used again, it will be invalidated but the second refresh token (of the second login) will stay valid and usable.

## Rate limiting
- All endpoints are rate limited by Auth0

## User account deletion

Users should contact the administrative people to delete their accounts.
  - There is no client endpoint for this action
  - Usually this can be done by exposing an endpoint from our server. Our server would then use the Auth0 administration API to delete the user. This usually fits a "web app+web server" scenario rather than a "native client + resource server" scenario. Plus, I am trying not to include any Auth0 secret in our server.

## Environments Architecure

auth0_arch

- We have two auth0 tenants: a production one and a development one. Each will have its own user database and its own API calls rate limit quota.
  - **Production tenant**: One "API" is registered with this tenant: `https://api.medperf.org` corresponding to the MedPerf main server.
  - **Development tenant**: One "API" is registered with this tenant:  `https://localhost:dev` corresponding to a local MedPerf server running for development/CI.
 
- MedPerf client will operate in three different environments:
  - **Production environment:** it can only communicate with a production MedPerf server and production auth0 tenant.
  -  **Development environment with online Auth:** it can only communicate with a local MedPerf server and the developement auth0 tenant. This is mainly used for integration tests with the auth provider and for developing auth-specific features. The local MedPerf server in this case should be spawned with the `.env.local.online-auth` env file (found in the `server` folder).
  - **Local Development environment:** it can only communicate with a local MedPerf server. Authorization of the MedPerf client to the local server in this settings will be using local tokens that can be found in the repository under the `mock_tokens` folder. The local MedPerf server in this case should be spawned with the `.env.local.local-auth`  env file (found in the `server` folder). 
 
- For the server integration tests (`seed.py`), it operates in two auth modes:
  -  `--auth local`: where it uses the local tokens to access the local MedPerf server.
  - `--auth online`: here it is configured to retrieve tokens from auth0 for the mock users that are created in the development tenant. Note that to retrieve the tokens here, we use the password-grant instead of the device authorization flow for simplicity. The code for this is available in `server/auth_provider_token.py`.

- For auth integration tests, the tests retrieve tokens for the mock users that are in the auth0 development tenant. For that, it uses a password from github secrets. For developers who have the password and want to run this locally, simply set the `MOCK_USERS_PASSWORD` environment variable prior to executing the tests.

## Django Admin

The super admin user will be a Django User. Through the admin panel, the super admin user will be able to give other users admin privileges. The super admin user will not be able to interact with the REST API. Other admin users will be able to do that, but will not be able to access the admin dashboard (they don't have a password).

### Misc

- Should we ID users in Django by `user_id` or by the tuple (`user_id`, issuer)? Is there a possibilty in the future to use multiple Auth0 issuers?
  - No
- If a user asked to delete their account, how can we assure they are the account owner? Auth0 doesn't send account deletion confirmation emails.
  - We will manually verify that.

## Opens

- Should we cache JWKs for longer on the client? on the server?


hasan7n:
TODOs:

- [ ] Client integration tests
- [ ] any unit tests?
- [ ] decide on how tutorials will run (can we guarantee that the selenium script will work on users machines?)
- [ ] implement logout
- [ ] client checks for email format and password strength before sending signup and password reset requests.
- [ ] add proper logging
- [ ] handle possible json decode error
- [ ] refactor+ code documentation

- [ ] decide about django admin user
- [ ] on django, fix `drf_spectacular` giving warnings after we used JWT authenticaton (something to do with API documentation)
- [ ] blacklisting access tokens in django (should we?)

- [ ] Manual testing


Manual testing:

- Signup:
  - [ ] invalid email format -> explicit error (should do client checks before sending)
  - [ ] weak/common passwords -> explicit error (should do client checks before sending)
  - [ ] existing user -> got the message: "invalid signup". (not explicit to prevent email enumeration)
- login:
  - [ ] blocked user -> explicit error
  - [ ] wrong password/email (happens on the UI) -> the code keeps running normally until multiple invalid attempts or expiration of request.
  - [ ] unverified email -> explicit error
- password change:
  - [ ] unregistered user -> 200 OK. (not explicit to prevent email enumeration)
- refresh token:
  - [ ] old refresh token -> error: Unknown or invalid refresh token.
  - [ ] new refesh token after attempted to use the old one -> error: Unknown or invalid refresh token.
  - [ ] blocked user -> explicit error: user is blocked
  - [ ] manually revoked token -> error: Unknown or invalid refresh token.
  - [ ] deleted user -> explicit error: The refresh_token was generated for a user who doesn't exist anymore.
- [ ] unverified users who reset their password became automatically verified
- [ ] Django decode a valid access token
- [ ] Django decode an access token with unexpected audience
- [ ] how errors due to rate limits will show up.

**NOTE**: login of unverified emails or blocked ones in the Auth0 web page UI will just say "activation failed"
but the json response printed by the medperf CLI will contain the specific reason of failure.

hasan7n:
Manual testing:

- Signup:
  - [x] invalid email format -> explicit error (should do client checks before sending)
  - [x] weak/common passwords -> explicit error (should do client checks before sending)
  - [x] existing user -> got the message: "invalid signup". (not explicit to prevent email enumeration)
- login:
  - [x] blocked user -> explicit error
  - [x] wrong password/email (happens on the UI) -> the code keeps running normally until multiple invalid attempts or expiration of request.
  - [x] unverified email -> explicit error
- password change:
  - [x] unregistered user -> 200 OK. (not explicit to prevent email enumeration)
- refresh token:
  - [x] old refresh token -> error: Unknown or invalid refresh token.
  - [x] new refesh token after attempted to use the old one -> error: Unknown or invalid refresh token.
  - [x] blocked user -> explicit error: user is blocked
  - [x] manually revoked token -> error: Unknown or invalid refresh token.
  - [x] deleted user -> explicit error: The refresh_token was generated for a user who doesn't exist anymore.
- [x] unverified users who reset their password became automatically verified
- [x] Django decode a valid access token
- [x] Django decode an access token with unexpected audience
- [ ] how errors due to rate limits will show up.

**NOTE**: login of unverified emails or blocked ones in the Auth0 web page UI will just say "activation failed"
but the json response printed by the medperf CLI will contain the specific reason of failure.

cli/medperf/commands/auth/login.py
msheller, line 1: I don't understand the last line. Why do we need to unset the user id?

cli/medperf/comms/auth/auth0.py
```diff
@@ -0,0 +1,247 @@
...
+        headers = {"content-type": "application/x-www-form-urlencoded"}
+        body = {
+            "grant_type": "urn:ietf:params:oauth:grant-type:device_code",
+            "device_code": device_code,
+            "client_id": self.client_id,
+        }
+
+        while True:
+            time.sleep(polling_interval)
+            issued_at = time.time()
```
msheller, line 133: Silly question: should this be before or after post?

cli/medperf/utils.py
```diff
@@ -55,36 +55,41 @@ def setup_logging(log_lvl):
 
 def delete_credentials():
     config_p = read_config()
```
msheller, line 57: I'd like to discuss this in our 6/29 meeting to be sure I am understanding it.

cli/medperf/utils.py
```diff
@@ -508,3 +525,30 @@ def get_cube_image_name(cube_path: str) -> str:
     except KeyError:
         msg = "The provided mlcube doesn't seem to be configured for singularity"
         raise MedperfException(msg)
+
+
+def validate_password(password):
+    # NOTE: if password checks are changed, make sure you modify config.password_policy_msg
```
msheller, line 531: Are these regexs taken from somewhere? if so, we should credit the source.

cli/medperf/comms/auth/auth0.py
hasan7n: I chose to record before, based on "better to refresh the token early than being late".

cli/medperf/utils.py
hasan7n: chatgpt

aristizabal95 (commented):
My review is not finished yet, but I'm submitting my current comments for discussion

.github/workflows/docker-ci.yml
```diff
@@ -19,18 +6,47 @@ jobs:
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.9'
+
+    - name: Setup Chrome
+      uses: browser-actions/setup-chrome@v1
+      with:
+        chrome-version: stable
+
+    - name: Install Singularity
```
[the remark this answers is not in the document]
aristizabal95: oh I see, let's open an issue then for reimplementing this on the future!

cli/medperf/tests/comms/test_auth0.py
```diff
@@ -0,0 +1,73 @@
...
+        "refresh_token": "",
+        "access_token": "",
+        "token_expires_in": 900,
+        "token_issued_at": time.time(),
+    }
+    mocker.patch(PATCH_AUTH.format("read_credentials"), return_value=creds)
+    spy = mocker.patch(PATCH_AUTH.format("Auth0._Auth0__refresh_access_token"))
+
+    # Act
+    Auth0().access_token
```
aristizabal95, line 41: If I understood correctly, retrieving this property should not be updated if the token hasn't expired. What should happen if it has expired? Maybe we should add a test case for that case as well

aristizabal95 (dismissed):
Hey Hasan, this looks amazing! Very organized and very thoughtful. I just left a few minor comments. I didn't review server changes

aristizabal95:
I removed my approval given that not everything is fully reviewed (server review is missing) and that it would be better for someone with more knowledge on the security concerns to approve this

PR #456, opened 2023-06-22
Fix result ls always filtering by benchmark

aristizabal95:
This PR fixes an issue where `medperf result ls` would always fetch results from a benchmark even if not passsed.


hasan7n:
Hey Alejandro! thanks! can you please rebase your changes according to the current main branch? I see a lot of old commits and also some extra changes related to another PR: #435

aristizabal95:
Huh, I thought I made sure I was working with the latest main branch. Thanks @hasan7n for reminding me! I must be messing something up with my workflow and I'll keep it in mind. It should look good now!

hasan7n (approved):
Looks good!

PR #459, opened 2023-06-28
GaNDLF tutorial fix

### LINKED ISSUE #436: [BUG] Documentation for creating a GaNDLF model
Following the [documentation of creating a GaNDLF model MLCube](https://docs.medperf.org/mlcubes/gandlf_mlcube/) works, but the resulting MLCube doesn't seem to work.

The root of the problem might be that the trained model itself, before being packaged as an MLCube, is faulty.

hasan7n:
This PR fixes some bugs faced when using MLCubes built according to [this guide](https://docs.medperf.org/mlcubes/gandlf_mlcube/). It also does some enhancements on the docs language.
Closes #436


aristizabal95 (approved):
LGTM

PR #473, opened 2023-08-03
Doc update

vpchung:
## Changelog
* Update docs: `medperf synapse_login` → `medperf auth synapse_login`


vpchung:
Note: the Synapse Python Client has [plans to deprecate](https://python-docs.synapse.org/build/html/news.html) non-auth token logins (e.g. username + password) as well as the `rememberMe` parameter by early 2024. I can file a follow-up ticket to update [this script](https://github.com/mlcommons/medperf/blob/main/cli/medperf/commands/auth/synapse_login.py)?

hasan7n:
@vpchung thanks for spotting this!
Regarding the deprecation of username-password auth, that sounds good! We can have a ticket to track this.

Issue #488, opened 2023-09-18
[BUG] Data folders are not mounted as read-only as expected


vpchung:
**Describe the bug**
During a recent benchmark run, I received an error about the data folder:

```
Traceback (most recent call last):
  File "/opt/conda/envs/medperf/lib/python3.9/site-packages/mlcube_docker/docker_run.py", line 233, in run
    mounts, task_args, mounts_opts = Shell.generate_mounts_and_args(
  File "/opt/conda/envs/medperf/lib/python3.9/site-packages/mlcube/shell.py", line 360, in generate_mounts_and_args
    _generate(params.inputs, IOType.INPUT)  # Process input parameters.
  File "/opt/conda/envs/medperf/lib/python3.9/site-packages/mlcube/shell.py", line 285, in _generate
    raise ConfigurationError(mlcube.errors.ConfigurationError: Invalid task: task=infer, param=data_path, type=unknown. Type is unknown and unable to identify it (/home/vchung/.medperf/api_medperf_org/data/7/data).
```

Digging further, it looks like the dataset was completely wiped of its data:

```
$ ll .medperf/api_medperf_org/data/7/
total 1
-rw-r--r--    1 vchung  staff   543B Sep  8 15:05 registration-info.yaml
```

My suspicion is that one of the previous benchmark runs did a cleanup that resulted in cleaning out the input folder (in this case, `.medperf/api_medperf_org/data/7/data` and `.medperf/api_medperf_org/data/7/labels`).  This would further indicate that the input folders are not being mounted as read-only as expected.

**To Reproduce**
@hasan7n has created a MRE.

**Expected behavior**
The data folders should remain unchanged during each container run.

**Desktop (please complete the following information):**
 - OS: Debian GNU/Linux 11
 - Browser: N/A
 - Version: MedPerf version 0.1.0

hasan7n:
Thanks @vpchung for reporting this!

#### Additional context
There is a typo in the passed MLCube parameters that make input folders and files read-only:
We use e.g.:
`Ptasks.evaluate.parameters.input.predictions.opts`

but it should be:

`-Ptasks.evaluate.parameters.inputs.predictions.opts`

(There was a missing dash `-` and a missing `s` in `inputs`)

Currently, MLCube ignores the parameters (because there is a typo) and hence doesn't enforce the inputs to be read-only.
+ we should have an integration test for such scenario.

PR #492, opened 2023-10-06
Fix mount permissions on input volumes

aristizabal95:
This PR fixes an simplifies ensuring read-only access to input volumes on MLCube execution. This is done so by leveraging on the recent `--mount=ro` flag, instead of specifying per-volume permissions


msheller (commented):
Two questions:

1. I think I am missing something. It looks to me like "*_str_params" isn't used in the run function in the old code. If so, wouldn't the fix have been to use them? Or was that variable intending to use a feature in MLCube that is broken/missing?

2. Am I correct that MLCube won't set 'ro' for the output directories, even though you've specified --mount=ro? If so, I wonder if we should request a name change on that parameter, as it's confusing to me to see "--mount" and it not apply to all mount points.

hasan7n (approved):
Looks good!

hasan7n:
@msheller 
1) It was used. The first line of the `cube.run` was updating the `kwargs` with the string parameters. The cause of the bug is described here: #488 
2) Yes, the `--mount` will only act on the inputs. I will open an issue for name change in the mlcube repo.

PR #493, opened 2023-10-10
Fix dataset hashes not taking labels into account

### LINKED ISSUE #507: Dataset hashes not taking labels into account
Details in the linked pull request

aristizabal95:
This small PR adds the required changes for generating dataset hashes from both the data and labels. Previously, only the data path was considered, which meant that two datasets with the same input data but different labels would register as the same dataset
closes #507


msheller (approved):
You know, I have never seen the approach of hashing each entry, then sorting those before combining them. Very interesting! I think it is analogous to the method I am used to, where you sort all the input values before hashing. If something hits me, I'll let you know, but I really don't see why this doesn't work well (and the implementation is cleaner).

Maybe add a comment though just to clarify that your hash is not depending on the ordering of the walk function calls? A new reader might see the lack of sorted(...) and immediately get concerned (which is what I did until I saw what you were doing instead :P).

Thanks!

hasan7n (approved):
Thanks!

PR #494, opened 2023-10-10
Fix MLCube configure silently failing

aristizabal95:
This PR ensures errors while configuring an MLCube are not ignored, and instead they raise an exception.


msheller (approved):
Nice and simple. Looks good.

PR #502, opened 2023-11-19
Server tests

### LINKED ISSUE #509: Implement server tests
Implement server tests

hasan7n:
This PR adds tests for all server endpoints, focusing on permissions / disallowed actions.

TODO: update [this](https://docs.google.com/spreadsheets/d/1Y-fugi1Y6swtUfC91W1zg3HAH2Yi4Xr7y_nGFlYH7HU/edit#gid=273022396). EDIT: done

closes #509


cli/medperf/comms/rest.py
```diff
@@ -183,8 +183,10 @@ def get_benchmark_models(self, benchmark_uid: int) -> List[int]:
         Returns:
             list[int]: List of model UIDS
         """
-        models = self.__get_list(f"{self.server_url}/benchmarks/{benchmark_uid}/models")
-        model_uids = [model["id"] for model in models]
+        # TODO: filter to find approved ones only if we decide to have
```
aristizabal95, line 186: Should we handle filtering within the comms layer? Or within the entity layer? I think we've done filtering on the entity layer so far, because we also need to apply filtering to local entities

hasan7n: It's tricky for associations since we don't have an association `Entity` on the client yet.

Filtering associations requires two steps:
1. first, finding the latest association of each mlcube. Ideally this should live in an association `Entity`, but for now I think this should live in the comms layer, since the client doesn't actually care about the history of associations. Also, this logic is required for all places in the client code where this comms function is called.
2. Then, ordinary filtering by approval status. Currently this filtering is required in two places:
  a. `entity.list` command: which already applies filtering in the commands layer.
  b. `benchmark.get_models_uids`: which by definition should return approved associations model IDs. We can apply filtering inside this function.

I think this will become cleaner if we decide to create an association `Entity` in our code base.

server/benchmark/models.py
```diff
@@ -20,7 +20,7 @@ class Benchmark(models.Model):
     description = models.CharField(max_length=100, blank=True)
     docs_url = models.CharField(max_length=100, blank=True)
     owner = models.ForeignKey(User, on_delete=models.PROTECT)
-    demo_dataset_tarball_url = models.CharField(max_length=256, blank=True)
+    demo_dataset_tarball_url = models.CharField(max_length=256)
```
aristizabal95, line 23: I remember we talked about this before, but can you remind me how we're getting past situations where demo datasets can't be shared directly with a link?

aristizabal95 (approved):
Hey Hasan! Just left a few comments/questions for the future. I skimmed over the server tests and implementation. Everything looks great and the tests look very well organized. Amazing amazing work as always Hasan!

server/benchmark/models.py
hasan7n: currently we support using synapse storage (e..g, Verena used a private demo dataset)
we should eventually support things like dropbox or google drive

cli/medperf/tests/entities/test_benchmark.py
```diff
@@ -20,31 +25,36 @@ def setup(request, mocker, comms, fs):
...
-    def test_benchmark_includes_reference_model_in_models(self, setup):
-        # Act
-        id = setup["remote"][0]
-        benchmark = Benchmark.get(id)
-
-        # Assert
-        assert benchmark.reference_model_mlcube in benchmark.models
-
-    def test_benchmark_includes_additional_models_in_models(self, setup):
+    def test_benchmark_get_models_works_as_expected(self, setup, expected_models):
```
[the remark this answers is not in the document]
hasan7n: that's a very valid point

PR #511, opened 2023-11-26
Store email address in medperf database

hasan7n:
In the near future we may implement sending email notifications. Access tokens should include a custom claim named `https://medperf.org/email` which will include the user's email address. This PR will store the user's email address in the medperf database the first time the user accesses the server


aristizabal95 (approved):
Looks great!

Issue #496, opened 2023-10-22
[FEATURE] Filter valid entities when using the `ls` command 


hasan7n:
`medperf <entity> ls` currently doesn't have an option to only show entities that have `is_valid` flag as `True`. This feature was requested from BraTS2023 organizers.

The solution would be to have an option (e.g. `medperf <entity> ls --valid`) that lists only valid entities.

aristizabal95:
This filtering is client-side

VukW:
merge master here and see what's been changed during this half of year

PR #513, opened 2023-11-27
Added filtering option for entities listing

VukW:
Solution for https://github.com/mlcommons/medperf/issues/496 : add `--valid` flag for entity listing:
```
medperf mlcube ls --valid
medperf dataset ls --valid
medperf benchmark ls --valid
medperf result ls --valid
```

Waits for https://github.com/mlcommons/medperf/issues/509 to implement `is_valid` flag for `Result` entity

### DIFF
--- a/cli/medperf/entities/interface.py
+++ b/cli/medperf/entities/interface.py
@@ -5,15 +5,14 @@ from abc import ABC, abstractmethod
 class Entity(ABC):
     @abstractmethod
     def all(
-        cls, local_only: bool = False, comms_func: callable = None
+        cls, local_only: bool = False, filters: dict = None
     ) -> List["Entity"]:
         """Gets a list of all instances of the respective entity.
-        Wether the list is local or remote depends on the implementation.
+        Whether the list is local or remote depends on the implementation.
 
         Args:
-            local_only (bool, optional): Wether to retrieve only local entities. Defaults to False.
-            comms_func (callable, optional): Function to use to retrieve remote entities.
-                If not provided, will use the default entrypoint.
+            local_only (bool, optional): Whether to retrieve only local entities. Defaults to False.
+            filters (dict, optional): key-value pairs specifying filters to apply to the list of entities.
 
         Returns:
             List[Entity]: a list of entities.
--- a/cli/medperf/entities/schemas.py
+++ b/cli/medperf/entities/schemas.py
@@ -107,5 +107,6 @@ class ApprovableSchema(BaseModel):
             status = Status(v)
         return status
 
+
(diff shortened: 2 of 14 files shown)


cli/medperf/entities/interface.py
```diff
@@ -5,15 +5,14 @@
 class Entity(ABC):
     @abstractmethod
     def all(
-        cls, local_only: bool = False, comms_func: callable = None
+        cls, local_only: bool = False, filters: dict = None
```
VukW, line 8: This fix was not connected to ls valid feature, just a bugfix: making abstract function signature relevant to a real implementations

cli/medperf/entities/schemas.py
```diff
@@ -105,3 +106,6 @@ def default_status(cls, v):
         if v is not None:
             status = Status(v)
         return status
+
+class DeployableEntity(DeployableSchema, Entity):
+    pass
```
VukW, line 111: This "implementation" is necessary at two places:
1. Entity list command - to declare that applicable objects should be both DeployableSchema and Entity.
2. for Entity list test purposes
However, placing implementation here breaks code design a bit: it make `schemas.py` dependable from `interface.py`. Maybe it's worth to move this class definition to other place?

cli/medperf/tests/commands/mlcube/test_list.py
```diff
@@ -0,0 +1,101 @@
...
+def cls_local_cubes(*args, **kwargs) -> list[Cube]:
+    return [
+        Cube(**generate_cube(id=101, is_valid=True, owner=1)),
+        Cube(**generate_cube(id=102, is_valid=False, owner=1)),
+        # Intended: for local mlcubes owner is never checked.
+        # All local cubes are supposed to be owned by current user
+
+        # generate_cube(id=103, is_valid=True, owner=12345),
+        # generate_cube(id=104, is_valid=False, owner=12345),
+    ]
```
VukW, line 32: The current behavior is not to filter local cubes by owner. I didn't change it, instead I anchored it in this test case. In future, once we elaborate & implement new `ls` behavior, these tests have to be changed

cli/medperf/entities/schemas.py
aristizabal95: I see your point. Maybe this could live in the interface file instead? I also think that, just for clarity, entities that apply to this description should inherit from this DeployableEntity class.

aristizabal95 (commented):
Looks good! I left a few comments. The only important change I would consider valuable is to the tests definition for the list method.

hasan7n:
closing as outdated

PR #521, opened 2024-01-12
Codespaces

VukW:
Added devconfig for running medperf in codespace.

1. raised PyYAML dependency to support py3.10
2. Neat point: the codespace is created once per user per branch, and `medperf` repo is attached only on the creation. Thus, the state of user's codespace is persistent despite of new commits to main branch. Therefore, if we commit anything, users have to recreate the codespace to get updated version.


hasan7n (dismissed):
Great work man!

hasan7n (approved):

PR #523, opened 2024-01-19
Add signal for SIGINT during mlcube call

aristizabal95:
This PR handles an error where if a user does CTRL+C during an MLCube call, the MLCube process doesn't close, leaving a dangling and compute-intesive process running in the background

### DIFF
--- a/cli/medperf/entities/cube.py
+++ b/cli/medperf/entities/cube.py
@@ -1,5 +1,6 @@
 import os
 import yaml
+from signal import signal, SIGINT
 import pexpect
 import logging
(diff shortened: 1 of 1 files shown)


hasan7n (changes requested):
Thanks Alejandro!

Did you test this? It looks like this handler will stay the handler even after mlcube execution has finished. Also it seems that interrupting the execution will allow medperf code to continue execution, which will result in bugs (e.g. result file hasn't been created)
also it doesn't respect if other parts of the code had already defined a handler (we are doing this for report submission)

PR #524, opened 2024-01-22
Mlcube download logging

VukW:
Catches `docker pull` output and put it directly to medperf's stdout:
image
The result is a bit weird, but at least it shows the `docker pull` is running and smth is executing

### DIFF
--- a/cli/medperf/entities/cube.py
+++ b/cli/medperf/entities/cube.py
@@ -8,6 +8,7 @@ from pathlib import Path
 
 from medperf.utils import (
     combine_proc_sp_text,
+    DebugAndOutputLogger,
     log_storage,
     remove_path,
     verify_hash,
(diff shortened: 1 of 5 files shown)


cli/medperf/entities/cube.py
```diff
@@ -220,13 +221,16 @@ def download_image(self):
             # Retrieve image from image registry
             logging.debug(f"Retrieving {self.id} image")
             cmd = f"mlcube configure --mlcube={self.cube_path}"
-            with pexpect.spawn(cmd, timeout=config.mlcube_configure_timeout) as proc:
-                proc_out = proc.read()
+            cmd_env = {**os.environ, 'DOCKER_CLI_HINTS': 'false'}
+            with pexpect.spawn(cmd, timeout=config.mlcube_configure_timeout, env=cmd_env) as proc:
+                proc.logfile_read = DebugAndOutputLogger()  # Redirect cmd output to medperf stdout
```
VukW, line 226: I was exploring to use `combine_proc_sp_text` there to display docker logs in a handy way. However, this function does not work with non-linear output (i.e. when output is modifying in the process with escape sequences), and I wasn't successful to fix that. Thus, offering the current, a few ugly, but still solution

cli/medperf/entities/cube.py
```diff
@@ -220,13 +221,16 @@ def download_image(self):
             # Retrieve image from image registry
             logging.debug(f"Retrieving {self.id} image")
             cmd = f"mlcube configure --mlcube={self.cube_path}"
-            with pexpect.spawn(cmd, timeout=config.mlcube_configure_timeout) as proc:
-                proc_out = proc.read()
+            cmd_env = {**os.environ, 'DOCKER_CLI_HINTS': 'false'}
```
VukW, line 224: this env hides `What's next?` docker CLI hits from its output, leaving only essential logs about pulling docker layers

VukW:
Prettified output according to guidelines
image

cli/medperf/utils.py
```diff
@@ -245,6 +246,27 @@ def combine_proc_sp_text(proc: spawn) -> str:
...
+    def __init__(self):
+        self._just_started_flag = True
+
+    def write(self, byte):
+        if self._just_started_flag:
+            sys.stdout.write("\r\033[K")  # clean the current spinner line
+            self._just_started_flag = False
+        s = byte.decode("utf-8")
+        sys.stdout.write(f'{Fore.WHITE}{Style.DIM}{s}{Style.RESET_ALL}')
+        sys.stdout.flush()
```
VukW, line 262: The `yasmin` spinner interferes too much into the printed lines to ensure statis text is always shown. That's why it's impossible to print dynamically changed logs with escape sequences via `config.ui.write()` and I had to print out everything manually via `sys.stdout`

hasan7n:
@VukW should we close this?

PR #526, opened 2024-01-30
Singularity support, refactor cube.download

hasan7n:
This PR:
- Singularity 3.10 is now supported
- clarify how GPUs are used 
- configuring the mlcube (e.g. pull docker image or converting it to singularity) now happens only when needed, not whenever the mlcube is retrieved

### DIFF
--- a/.github/workflows/docker-ci.yml
+++ b/.github/workflows/docker-ci.yml
@@ -18,6 +18,7 @@ jobs:
         python-version: '3.9'
 
     - name: Install Singularity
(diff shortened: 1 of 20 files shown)


aristizabal95 (approved):
Looks good! It would be great if we write a document where we keep track of all the runner-specific logic we've been adding here and the reasons for them. We could share that with the Best Practices team to have discussions and determine if there's a way to incorporate that logic on the MLCube instead

PR #527, opened 2024-02-01
Debug logging

VukW:
- [x] change medperf default log level to DEBUG
- [x] set mlcube loglevel to DEBUG (and print debug & info logs to debug file only)
- [x] `--container-loglevel` param: is passed to the running containers (Docker engine only)
- [ ] graceful closing child processes on keyboard interrupt (via `with` statement)
- [x] parse mlcube framework output to redirect its debug logs to debug file only, not to stdout
- [x] Add test runs for debug env variable
- [ ] ~~check medperf debug log calls, assure we cover all the necessary of medperf parts / functions while debugging~~ Won't be done right now
- [ ] ~~Support singularity in setting up container loglevel~~ Won't be done right now


cli/medperf/utils.py
```diff
@@ -228,6 +228,12 @@ def combine_proc_sp_text(proc: spawn) -> str:
     static_text = ui.text
     proc_out = ""
     break_ = False
+    log_pattern = re.compile(
+        r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2} \S+ \S+\[(\d+)\] (DEBUG|INFO) (.*)$"
```
hasan7n, line 232: we should only hide DEBUG logs. INFO should be shown

cli/medperf/utils.py
```diff
@@ -237,23 +243,39 @@ def combine_proc_sp_text(proc: spawn) -> str:
...
+        line_should_be_filtered = False
+        if match:
+            line_pid, matched_log_level_str, content = match.groups()
+            matched_log_level = logging.getLevelName(matched_log_level_str)
+
+            # if line matches conditions, it is just logged to debug; else, shown to user
+            line_should_be_filtered = (line_pid == str(proc.pid)  # hide only `mlcube` framework logs
+                                       and isinstance(matched_log_level, int)
+                                       and matched_log_level < logging.WARNING)  # hide only debug and info logs
+
```
hasan7n, line 260: why are we checking the matched log level here? Shouldn't this be already taken care of in the regex?

cli/medperf/utils.py
```diff
@@ -237,23 +243,39 @@ def combine_proc_sp_text(proc: spawn) -> str:
             logging.error("Process timed out")
             raise ExecutionError("Process timed out")
         line = line.decode("utf-8", "ignore")
-        if line:
+
+        if not line:
```
hasan7n, line 247: I suggest we have the logic of "deciding if we want to filter" be encapsulated in a separate function

cli/medperf/utils.py
```diff
@@ -237,23 +243,39 @@ def combine_proc_sp_text(proc: spawn) -> str:
...
+
+        match = log_pattern.match(ansi_escape_pattern.sub('', line))
+        line_should_be_filtered = False
+        if match:
+            line_pid, matched_log_level_str, content = match.groups()
+            matched_log_level = logging.getLevelName(matched_log_level_str)
+
+            # if line matches conditions, it is just logged to debug; else, shown to user
+            line_should_be_filtered = (line_pid == str(proc.pid)  # hide only `mlcube` framework logs
+                                       and isinstance(matched_log_level, int)
```
hasan7n, line 258: I suggest we have the logic of "deciding if we want to filter" be encapsulated in a separate function

cli/medperf/ui/cli.py
```diff
@@ -39,6 +41,7 @@ def print_warning(self, msg: str):
         self.__print(msg)
 
     def __print(self, msg: str = ""):
+        logging.info(msg)
```
hasan7n, line 44:
```suggestion
        logging.info("Console: " + str(msg))
```

cli/medperf/entities/cube.py
```diff
@@ -355,9 +363,9 @@ def run(
         cmd += " -Pplatform.accelerator_count=0"
 
         logging.info(f"Running MLCube command: {cmd}")
-        proc = pexpect.spawn(cmd, timeout=timeout)
-        proc_out = combine_proc_sp_text(proc)
-        proc.close()
+        with pexpect.spawn(cmd, timeout=timeout) as proc:
+            proc_out = combine_proc_sp_text(proc)
+
         if output_logs is None:
```
hasan7n, line 369: This doesn't seem to do the trick
pressing CTRL+C is not stopping the child process

cli/medperf/ui/cli.py
VukW: Btw about sensitive information. In debug logs we do already have a listed tree of `dataset` containment. May there be some sensitive information also?

cli/medperf/utils.py
VukW: Yep, you are right, thanks, that smelling code is a result of mixing two different approaches and I forgot to fix it.
Originally I used hardcoded level names in regex, but later decided to extract the loglevel and compare its severeness explicitly.

Fixed that by changing regexp: removed hardcoded loglevels from there

cli/medperf/utils.py
VukW: Agree, fixed

cli/medperf/utils.py
VukW: Ah, I see, you changed the default mlcube log level
Fixed 👌

cli/medperf/entities/cube.py
VukW: I see you are totally right.
When `medperf` -> `mlcube` -> `container` stack is active, stopping (ctrl+C) medperf stops `mlcube`, but does not stop container. Though when mlcube is running directly, stopping mlcube stops container also. 
I'll debug it deeper...

cli/medperf/ui/cli.py
VukW: Removed logging for now

cli/medperf/entities/cube.py
VukW: A separate issue for solving this: https://github.com/mlcommons/medperf/issues/531

hasan7n (approved):
Awesome work man! Thanks!

PR #532, opened 2024-02-15
thread-safe token refresh

hasan7n:
wraps the authentication class's `access_token` method under a threading lock to make it thread-safe and prevents the possibility of having multiple threads trying to refresh the access token, which may result in refresh token reuse since we use refresh token rotation


VukW:
LGTM 👍

VukW (approved):

PR #534, opened 2024-02-16
Killing all descendant processes on exception

VukW:
If medperf process (or mlcube, its ancestor) is stopped by any reasons, kills all children processes, stops docker containers, etc.

### DIFF
--- a/cli/medperf/entities/cube.py
+++ b/cli/medperf/entities/cube.py
@@ -1,6 +1,5 @@
 import os
 import yaml
-import pexpect
 import logging
 from typing import List, Dict, Optional, Union
 from pydantic import Field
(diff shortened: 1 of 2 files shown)


hasan7n (approved):
Looks great Slava! Nice work.
One small thing: do you know if [os.killpg](https://docs.python.org/3/library/os.html#os.killpg) works on MAC OS?
perhaps @aristizabal95 can you please test this on your machine?

VukW:
I don't have exact information, but it looks like it should be.
1. `os` is designed to support all `unix` (including `MacOS`) and `windows`
systems
2. For some functions documentation explicitly says "it is supported on
Unix except MacOS" https://docs.python.org/3.10/library/os.html#os.fdatasync
.There is no such notes on `killpg`
https://docs.python.org/3.10/library/os.html#os.killpg

So, I believe, it should work well

hasan7n (approved):

PR #535, opened 2024-02-18
Fix token storage bug

hasan7n:
- Drops the usage of `os.open` and uses the built-in `open` for storing the tokens. It creates an empty file first, changes its permissions, then appends the token.

This PR also Forces logout for authenticated users who try to login again (to prevent refresh token count build up; to not lose track of active refresh tokens without revoking them). This is not related to the storage bug, but it is better to be implemented.


VukW (dismissed):
LGTM, concise and pretty solution 

Right now setting tokens looks not so atomic, but in reality any solution of writing data to the file is never totally atomic. So, instead of bothering about protection there it would be much better to ensure that we read token safely & can rewrite tokens successfully if anything is broken with them (if file is empty, not exists, token is broken, you name it)

VukW (approved):

PR #541, opened 2024-02-18
mlcube logging small fixes

hasan7n:


VukW (dismissed):
As this PR solves the biggest part of issue, displaying to user that medperf is doing some he (though without realtime progress for docker image downloading), I believe, we can merge it as-is. The latter improvement https://github.com/hasan7n/medperf/pull/3 can be merged later after thorough check

aristizabal95 (approved):

PR #545, opened 2024-02-23
Mlcube download log

VukW:
Replacement of https://github.com/hasan7n/medperf/pull/3 as that base branch was already merged into main

While reading output from spawned mlcube process, stop&display not only when a full line (ending with \n) is found, but when a line update (ending with \r) is got.


hasan7n:
closing as outdated

PR #546, opened 2024-02-23
Mlcube download docker cli hints

VukW:
Replacement of https://github.com/hasan7n/medperf/pull/4 as that base branch was already merged into main.

Hide hints from docker cli output when we do `docker pull`


VukW:
Closing this PR as this fix is better to be done on mlcube side.

PR #550, opened 2024-02-27
package installation in non-editable mode: `pip install ./cli`

VukW:
The goal is to make medperf client installable from pypi. This PR is the first part that allows installing package not from repository

A few changes were implemented:

- Totally removed git branch check updates: as client installed in non-editable mode would not have any connections with source folder where it was installed from, it knows nothing about git branch.
- During deployment, local server places `cert.crt`, `cert.key` and `mock_tokens/tokens.json` to `~/.medperf_config/.local_server/` so detached medperf client can use it. NB: this change breaks the behavior for existing developers (as path to local server cert is already defined in user config). You should update local config's `certificate` path to `~/.medperf_config/.local_server/cert.crt`.
- `pip install -e ./cli` replaced with `pip install ./cli` everywhere


cli/medperf/config.py
```diff
@@ -1,16 +1,13 @@
 from ._version import __version__
 from pathlib import Path
 
-BASE_DIR = Path(__file__).resolve().parent.parent.parent
```
VukW, line 4: `BASE_DIR` has no sense when client is installed in standard way (as it points somewhere to `/your/path/to/python/env/lib/site-packages/`)

VukW:
Folks, I'm thinking. Do we need to pack local server to the pip package at all? I mean, it is used only for development purposes. What do you think if we leave it in a repo? so every developer who needs local server would have to check out repo.

VukW:
TODO:
- return git update check
- fetch only current branch instead of all of them?

PR #553, opened 2024-02-29
Fix caching bug in resources.download

### LINKED ISSUE #549: [BUG] MLCube assets are not updated after content or url change
**Describe the bug**
MLCubes are constructed from multiple assets, all of which are downloaded the first time such MLCube is requested by the CLI. After subsequent requests to the MLCube, cached assets are preferred. This is okay as long as the contents of the remote MLCube don't change. But, if assets are updated, contents are modified or the hash changes, while the MLCube metadata gets updated, assets are not redownloaded, or rechecked with the new hash. This leads to people running outdated MLCubes without knowledge, and can also lead to broken MLCube registration updates, because tests on the user side will pass due to local cache being valid instead of remote contents being revised.

**To Reproduce**
1. Submit an MLCube and run it
2. Change the URL of one of the assets (MLCube URL for example) to something nonsensical through the admin dashboard
3. Run the MLCube again. It runs as usual
4. Change the hash of one of the assets (parameters hash for example) to something nonsensical through the admin dashboard
5. Run the MLCube again. It still runs like normal
6. Remove mepderf's cache of that MLCube
7. Run the MLCube again. It now fails.

**Expected behavior**
When the asset url or hash changes, the local cache should be invalidated or removed before proceeding with the download.

**Additional context**
This raises a question. We can detect asset change by hash change or url change. What happens when the contents of the asset change, but not the url? We cannot detect content changes without downloading the asset again. Someone could submit an MLCube with, for example, a url for the latest mlcube.yaml in a repository. In that case, contents may change, but the link is the same. Users that have downloaded that MLCube previously will not have any issues later on because local cache is used for that asset, but new users of that MLCube will encounter issues because of hash mismatch. How can we proceed here? Should we prefer redownloading lightweight assets like yaml files? What about heavy assets like additional files?

hasan7n:
Details of the problem can be found in the issue this PR fixes.
Closes #549

### DIFF
--- a/cli/medperf/comms/entity_resources/resources.py
+++ b/cli/medperf/comms/entity_resources/resources.py
@@ -17,13 +17,23 @@ import os
 import logging
 import yaml
 import medperf.config as config
-from medperf.utils import generate_tmp_path, get_cube_image_name, remove_path, untar
-from .utils import download_resource, verify_or_get_hash
+from medperf.utils import (
+    generate_tmp_path,
+    get_cube_image_name,
+    remove_path,
+    untar,
+    get_file_hash,
+)
+from .utils import download_resource
 
 
 def _should_get_regular_file(output_path, expected_hash):
     if os.path.exists(output_path) and expected_hash:
-        if verify_or_get_hash(output_path, expected_hash):
+        calculated_hash = get_file_hash(output_path)
+        logging.debug(
+            f"{output_path}: Expected {expected_hash}, found {calculated_hash}."
+        )
+        if expected_hash == calculated_hash:
             logging.debug(f"{output_path} exists and is up to date")
             return False
         logging.debug(f"{output_path} exists but is out of date")
--- a/cli/medperf/comms/entity_resources/resources.py
+++ b/cli/medperf/comms/entity_resources/resources.py
@@ -165,7 +165,7 @@ def get_cube_additional(
     ):
         return expected_tarball_hash
 
-    # Download the additional files.Make sure files are extracted in tmp storage
+    # Download the additional files. Make sure files are extracted in tmp storage
     # to avoid any clutter objects if uncompression fails for some reason.
     tmp_output_folder = generate_tmp_path()
     output_tarball_path = os.path.join(tmp_output_folder, config.tarball_filename)
--- a/cli/medperf/comms/entity_resources/resources.py
+++ b/cli/medperf/comms/entity_resources/resources.py
@@ -21,18 +21,13 @@ from medperf.utils import generate_tmp_path, get_cube_image_name, remove_path, u
 from .utils import download_resource, verify_or_get_hash
 
 
-def _should_get_cube(output_path, expected_hash):
+def _should_get_regular_file(output_path, expected_hash):
     if os.path.exists(output_path) and expected_hash:
         if verify_or_get_hash(output_path, expected_hash):
             logging.debug(f"{output_path} exists and is up to date")
             return False
         logging.debug(f"{output_path} exists but is out of date")
     return True
-
-
-def _should_get_cube_params(output_path, expected_hash):
-    # Happens to be the same logic
-    return _should_get_cube(output_path, expected_hash)
 
 
 def _should_get_cube_additional(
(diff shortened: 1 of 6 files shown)


cli/medperf/comms/entity_resources/utils.py
```diff
@@ -60,10 +61,14 @@ def verify_or_get_hash(tmp_output_path: str, expected_hash: str) -> str:
...
+        str: Calculated hash of the asset. Returns None if hashes mismatch
     """
     calculated_hash = get_file_hash(tmp_output_path)
-    verify_hash(calculated_hash, expected_hash)
+    if expected_hash and calculated_hash != expected_hash:
+        logging.debug(
+            f"{tmp_output_path}: Expected {expected_hash}, found {calculated_hash}."
+        )
+        return
     return calculated_hash
```
VukW, line 72: 1. right now the function returns not str, but Optional[str]
But before fixing, look plz to the next comment at `download_resource`. This function is too complex now as it has different behavior depending on if `expected_hash` was passed; moreover, the function can be removed totally

cli/medperf/comms/entity_resources/utils.py
```diff
@@ -97,6 +102,9 @@ def download_resource(
 
     calculated_hash = verify_or_get_hash(tmp_output_path, expected_hash)
 
+    if calculated_hash is None:
+        raise InvalidEntityError(f"Hash mismatch: {resource}")
```
VukW, line 106: `verify_or_get_hash` right now has quite a complex behavior depending on if `expected_hash` was passed:
- if passed, then function returns either `None` or `expected_hash` 
- if not passed, then returns `calculated_hash`

Thus, you have to split hash comparison logic onto two parts, here and inside `verify_or_get_hash`. What do you think if we simplify it, removing `verify_or_get_hash` totally:

```suggestion
    calculated_hash = get_file_hash(tmp_output_path)

    if expected_hash and expected_hash != calculated_hash:
        raise InvalidEntityError(f"Hash mismatch: {resource}, found: {calculated_hash}, expected: {expected_hash}")
```

VukW: btw in which cases there may be no expected hash? Looks like it should always exist, isn't it?

cli/medperf/comms/entity_resources/resources.py
```diff
@@ -14,9 +14,46 @@
...
+import yaml
 import medperf.config as config
 from medperf.utils import generate_tmp_path, get_cube_image_name, remove_path, untar
-from .utils import download_resource
+from .utils import download_resource, verify_or_get_hash
+
+
+def _should_get_cube(output_path, expected_hash):
+    if os.path.exists(output_path) and expected_hash:
+        if verify_or_get_hash(output_path, expected_hash):
```
VukW, line 26:
```suggestion
        if expected_hash != get_file_hash(output_path):
```
Removing `verify_or_get_hash` with its complex logic, making debugging simpler

cli/medperf/comms/entity_resources/resources.py
```diff
@@ -14,9 +14,46 @@
...
+            logging.debug(f"{output_path} exists and is up to date")
+            return False
+        logging.debug(f"{output_path} exists but is out of date")
```
VukW, line 29: Another possible option:
```suggestion
        calculated_hash = get_file_hash(output_path)
        logging.debug(f"output_path} exists, {expected_hash=}, {calculated_hash=}")
        return expected_hash != calculated_hash
```

cli/medperf/comms/entity_resources/resources.py
```diff
@@ -118,21 +159,32 @@ def get_cube_additional(
...
     tmp_output_folder = generate_tmp_path()
     output_tarball_path = os.path.join(tmp_output_folder, config.tarball_filename)
     tarball_hash = download_resource(url, output_tarball_path, expected_tarball_hash)
 
     untar(output_tarball_path)
     parent_folder = os.path.dirname(os.path.normpath(additional_files_folder))
     os.makedirs(parent_folder, exist_ok=True)
+    if os.path.exists(additional_files_folder):
+        # handle the possibility of having clutter uncompressed files
+        remove_path(additional_files_folder)
```
VukW, line 179: 👍

cli/medperf/comms/entity_resources/resources.py
```diff
@@ -118,21 +159,32 @@ def get_cube_additional(
...
-    if os.path.exists(additional_files_folder):
+    mlcube_cache_file = os.path.join(cube_path, config.mlcube_cache_file)
+    if not _should_get_cube_additional(
+        additional_files_folder, expected_tarball_hash, mlcube_cache_file
+    ):
         return expected_tarball_hash
 
-    # make sure files are uncompressed while in tmp storage, to avoid any clutter
-    # objects if uncompression fails for some reason.
+    # Download the additional files.Make sure files are extracted in tmp storage
```
VukW, line 168:
```suggestion
    # Download the additional files. Make sure files are extracted in tmp storage
```

VukW (dismissed):
Thanks for the fix, man, looks good! The only my comments are connected to `verify_or_get_hash` function that became quite complex and useless after refactoring, so, I believe we may just remove it totally

hasan7n:
@VukW I agree! Let me remove the function.
expected hash could be None if this was used during mlcube/benchmark submission while the user didn't provide the hash of the associated assets. Medperf client in this case will calculate the hash of the downloaded asset and considers it as the ground truth hash.

cli/medperf/comms/entity_resources/resources.py
aristizabal95: I do prefer splitting that logic, @hasan7n would something like this cause issues?

cli/medperf/comms/entity_resources/resources.py
```diff
@@ -14,9 +14,46 @@
...
+        if verify_or_get_hash(output_path, expected_hash):
+            logging.debug(f"{output_path} exists and is up to date")
+            return False
+        logging.debug(f"{output_path} exists but is out of date")
+    return True
+
+
+def _should_get_cube_params(output_path, expected_hash):
+    # Happens to be the same logic
+    return _should_get_cube(output_path, expected_hash)
```
aristizabal95, line 35: Might be pedantic, but in this case we should have a generic name for this function (`should_get_file` or `should_get_asset`?) and have both `_should_get_cube` and `_should_get_cube_params` call that. It makes it just a bit more readable.

cli/medperf/comms/entity_resources/resources.py
```diff
@@ -53,8 +92,10 @@ def get_cube_params(url: str, cube_path: str, expected_hash: str = None) -> str:
         hash_value (str): The hash of the downloaded file
     """
     output_path = os.path.join(cube_path, config.workspace_path, config.params_filename)
-    if os.path.exists(output_path):
+    if not _should_get_cube_params(output_path, expected_hash):
         return output_path, expected_hash
+    if os.path.exists(output_path):
+        remove_path(output_path)
```
aristizabal95, line 98: This function and `get_cube` are pretty similar, can we DRY up this code?

cli/medperf/comms/entity_resources/utils.py
aristizabal95: Agree with this sentiment

aristizabal95 (changes requested):
It's looking great! I do agree with Slava's points, and think a little bit of DRYing up can be done as well.

VukW (approved):

aristizabal95 (approved):
Looks great! Thanks for integrating the suggestions!

PR #555, opened 2024-03-14
Better debugging experience - Host info & packaging

aristizabal95:
This PR adds more information about the host machine in the logs, as well as provide a helper script to package all logs for sharing to the dev team. Lastly, the number of logs has been increased to cover a larger amount of commands executed by the user.

A few open questions:
- Should we log this as "info" or "debug"? Currently is set as info so that all logs contain it, but it feels more like the purpose of this information is for debugging.
- Currently, packaging logs is a separate script. Should we add this as a medperf command?
- User information regarding sudo and docker group doesn't seem to work on mac, but it may be because of a possible difference in how mac and linux handle this.


hasan7n (changes requested):
Nice work!
- Let's move the new functions in `utils.py` under the `logging` folder? Our utils file is already in pain :). Maybe we can create a `utils.py` file under the `logging` folder...
- I think the logs should be in `debug` mode. But anyway, I think at some point we should remove the verbosity options of logging into files, and let the verbosity option control how things are logged to the screen, not to files.
- Regarding the external script of packaging: it's not necessary but let's have this automatically packaged at the end of each command?

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,274 @@
...
+        "Total Memory": memory.total,
+        "Available Memory": memory.available,
+        "Used Memory": memory.used,
+        "Memory Usage Percentage": memory.percent,
+    }
+
+
+def get_disk_usage():
+    # Get disk usage
+    disk_usage = psutil.disk_usage("/")
```
hasan7n, line 47: perhaps also run this for $HOME and for medperf storage location?

hasan7n: actually we don't need it for `/`

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,274 @@
...
+        "Total Disk Space": disk_usage.total,
+        "Used Disk Space": disk_usage.used,
+        "Free Disk Space": disk_usage.free,
+        "Disk Usage Percentage": disk_usage.percent,
+    }
+
+
+def get_user_information():
+    # Get user information
+    username = getpass.getuser()
```
hasan7n, line 58: https://docs.python.org/3/library/getpass.html#getpass.getuser
possible exceptions could be raised

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,274 @@
...
+    is_sudoers = "sudo" in [g.gr_name for g in grp.getgrall() if username in g.gr_mem]
```
hasan7n, line 59: is there a way without iterating over all groups in the system? I think there should be a way without iterating over all groups in the whole system.

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,274 @@
...
+                "--format=csv",
+            ],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
+        output, _ = p.communicate()
+        if p.returncode != 0:
+            raise ExecutionError("nvidia-smi not installed/available")
+        gpus_data = [row for row in csv.DictReader(output.split("\n"))]
```
hasan7n, line 212: what if the output contained a warning message? this may fail right

hasan7n (changes requested):
I left a few comments on the functions used to retrieve the information.
Please check the comments and also please recheck the functions for possible exceptions.

I suggest to not use python APIs as much as possible (they are anyway not platform-agnostic). Just make sure commands work in UNIX systems. (e.g., just run `id` for user information).
Let's run commands through `subprocess` and just catch the stdout+stderr, without even parsing them (e.g., no need to parse nvidia-smi output, just dump it in the logs). The aim is to avoid exceptions (warning messages always can screw things up). What do you think?

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,274 @@
...
+            try:
+                value = str(value)
+            except Exception:
+                value = "<OBJECT>"
+        if isinstance(value, dict):
+            value = filter_var_dict_for_yaml(value)
+        filtered_dict[key] = value
+
+    return filtered_dict
+
```
hasan7n, line 141: let's just dump the `config.yaml` file in the logs instead of reading `config`?

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,274 @@
...
+        g.gr_name for g in grp.getgrall() if username in g.gr_mem
+    ]
+    return {
+        "Username": username,
+        "Is in Sudoers Group": is_sudoers,
+        "Is in Docker Group": is_docker_group,
+    }
+
+
+def get_docker_information():
```
hasan7n, line 70: we also need to check `nvidia-container-cli`

cli/medperf/logging/utils.py
aristizabal95: I thought reading the config object would also provide updated values for both the profile and any passed configuraiton argument for a given command. Do you think there would be an issue in using the config object instead of reading the config.yaml file?

cli/medperf/logging/utils.py
aristizabal95: I'm now using nvidia-smi's -f parameter to store the query in a csv file and I'm parsing that file then. Would that work, or would you prefer just dumping the contents?

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,189 @@
...
+    sh_script = os.path.join(os.path.dirname(__file__), "get_host_info.sh")
+    p = subprocess.Popen(
+        ["bash", sh_script],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    output, _ = p.communicate()
+    if p.returncode != 0:
+        return "Could not execute bash script for additional debug information"
```
hasan7n, line 148: let's have the output as is even if an error occured. This will be more helpful right?

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def log_machine_details():
+    system_info = {}
+    system_info["System Info"] = get_system_information()
+    system_info["Memory Usage"] = get_memory_usage()
+    system_info["Disk Usage"] = get_disk_usage()
+    system_info["Medperf Configuration"] = get_configuration_variables()
+    system_info["Medperf Storage Contents"] = get_storage_contents()
+    system_info["Python Environment"] = get_python_environment_information()
```
hasan7n, line 160: let's wrap each line with a try-except block?

cli/medperf/logging/utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+    debug_dict = {"Machine Details": system_info}
+
+    logging.debug(yaml.dump(debug_dict, default_flow_style=False))
+    logging.debug(get_additional_information())
```
hasan7n, line 165: and wrap this line as well with a try except block

cli/medperf/logging/formatters/newline_formatter.py
```diff
@@ -0,0 +1,21 @@
...
+        """
+        Override format function
+        """
+        msg = logging.Formatter.format(self, record)
+
+        if record.message != "":
+            parts = msg.split(record.message)
+            msg = msg.replace("\n", "\n" + parts[0])
+
+        return msg
```
hasan7n, line 21: why are we overriding`__init__`?
Also, in both `__init__` and `format`, is `logging.Formatter` the same as `super()`?

cli/medperf/logging/formatters/newline_formatter.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def format(self, record):
+        """
+        Override format function
+        """
+        msg = logging.Formatter.format(self, record)
+
+        if record.message != "":
+            parts = msg.split(record.message)
+            msg = msg.replace("\n", "\n" + parts[0])
```
hasan7n, line 19: what does this do (I am a bit confused about splitting using the record.message)

cli/medperf/logging/formatters/newline_formatter.py
aristizabal95: You're right. This was taken directly from a Stack Overflow thread, but could be improved by the points you raised. I'll modify the code to remove or simplify unnecessary sections.

cli/medperf/logging/formatters/newline_formatter.py
aristizabal95: I'm adding comments to explain this. Basically we're separating the logging prefix (timestamp | section_of_code_calling_logging | loglevel) from the actual message so that it can applied to each newline.

This was necessary as these sections of logging were breaking the structure of the logging file due to newlines

hasan7n (approved):
Super!

PR #557, opened 2024-03-19
Update mlcube run error UX

hasan7n:
Updates MLCube version to remove unnecessary error log, to make sanity checks UI exists better. This is a hotfix until we figure out  a better solution on how sanity checks task should exit.

### DIFF
--- a/cli/medperf/decorators.py
+++ b/cli/medperf/decorators.py
@@ -1,4 +1,3 @@
-import os
 import sys
 import typer
 import logging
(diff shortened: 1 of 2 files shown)


hasan7n:
@aristizabal95 also I removed the msg that asks users to check logs. Better not ask them, they may get annoyed if they think they are supposed to look at the logs, while the logs are for us, full of debug logs msgs.

aristizabal95 (approved):
Looks good!

PR #561, opened 2024-03-25
Send exec status independent of report permissions

aristizabal95:
status: waiting for experiment leads to decide on wether this is OK


hasan7n:
status: waiting for experiment leads to decide on weather this is OK

cli/medperf/commands/dataset/prepare.py
```diff
@@ -313,8 +310,12 @@ def prompt_for_report_sending_approval(self):
         dict_pretty_print(example)
 
         msg = (
-            " \nDo you approve the automatic submission of summaries similar to the one above"
-            + " to the MedPerf Server throughout the preparation process?[Y/n]"
+            "\nYou can decide wether or not to send information about your dataset preparation"
```
VukW, line 313:
```suggestion
            "\nYou can decide whether or not to send information about your dataset preparation"
```

cli/medperf/commands/dataset/prepare.py
```diff
@@ -190,17 +189,15 @@ def run_prepare(self):
...
             # Inform the server that the process is interrupted
-            if self.allow_sending_reports:
-                report_sender.stop("interrupted")
-            raise
+            report_sender.stop("interrupted")
+            raise e
 
         self.ui.print("> Cube execution complete")
-        if self.allow_sending_reports:
+        if not self.dataset.for_test:
```
VukW, line 200: It seems you don't need this check anymore. `report_sender` expected to work well both with real data and test

VukW: Can you check plz one more time if everything between `report_sender.start/stop/*` and `_send_data()` works well for test datasets (as I believe you are more experienced in this piece of code). I believe it works well, just to explicitly ensure

cli/medperf/tests/commands/dataset/test_prepare.py
```diff
@@ -54,37 +54,58 @@ def test_get_prep_cube_downloads_cube_file(mocker, data_preparation, cube):
...
+    mocker, data_preparation, allow_sending_reports, cube, comms, fs
 ):
     # Arrange
     data_preparation.allow_sending_reports = allow_sending_reports
     mocker.patch.object(cube, "run")
-    start_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.start"))
-    stop_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.stop"))
+    mocker.patch.object(data_preparation.dataset, "write")
+    mocked_obs = mocker.create_autospec(spec=Observer)
+    mocker.patch(PATCH_REGISTER.format("Observer", side_effect=mocked_obs))
```
VukW, line 88: Isn't a typo here?
```suggestion
    mocker.patch(PATCH_REGISTER.format("Observer"), side_effect=mocked_obs)
```

VukW: It seems that if test is run this typo should break the test, shouldn't it? So if it didn't break an execution, can you check plz the test is really run

VukW (dismissed):
Looks good to me. Let's fix typos and run tests again, if it goes smoothly, then PR is ready to be merged

cli/medperf/commands/dataset/prepare.py
aristizabal95: Good point, I'm removing this check. Also everything looks fine with the report_sender methods, so I think we're good there!

VukW (approved):
LGTM

PR #564, opened 2024-03-27
Allow MLCube to pass details on missing/extra modalities

aristizabal95:
This is a simple PR that aligns the report generation on the side of RANO's Data Preparation MLCube to the stages being used by the monitoring tool. Now, The Data Preparation MLCube provides a helpful message when someone encounters the missing or extra modalities error.


hasan7n (approved):
Just a heads-up: some `comment` keys in this file will evaluate to an empty string, and some of them will evaluate to `None`

PR #570, opened 2024-04-01
Fix compatibility tests bug with local mlcubes

hasan7n:
In compatibility tests, we allow the usage of local mlcubes (i.e., a path to a local folder containing an mlcube.yaml file). The client in this case will create a mock mlcube metadata file, specifying a mock URL corresponding to the existing `mlcube.yaml` file. 
Until we  merged https://github.com/mlcommons/medperf/pull/553, the code was not checking hashes of existing resources (like mlcube.yaml). After merging that PR, the code checks the hash, finds it invalid, then tries to download the file, causing it to attempt to use `mock_url`.
This PR fixes the issue by having the client calculate and store the hash of the existing mlcube.yaml file during the preparation of the mock mlcube metadata file.

### DIFF
--- a/cli/medperf/commands/compatibility_test/utils.py
+++ b/cli/medperf/commands/compatibility_test/utils.py
@@ -1,6 +1,6 @@
 from medperf.commands.dataset.prepare import DataPreparation
 from medperf.commands.dataset.submit import DataCreation
-from medperf.utils import get_folders_hash, remove_path
+from medperf.utils import get_file_hash, get_folders_hash, remove_path
 from medperf.exceptions import InvalidArgumentError, InvalidEntityError
 
 from medperf.comms.entity_resources import resources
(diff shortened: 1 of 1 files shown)


cli/medperf/commands/compatibility_test/utils.py
```diff
@@ -48,11 +48,13 @@ def prepare_local_cube(path):
...
     if not os.path.exists(cube_metadata_file):
+        mlcube_yaml_path = os.path.join(path, config.cube_filename)
+        mlcube_yaml_hash = get_file_hash(mlcube_yaml_path)
         temp_metadata = {
             "id": None,
             "name": temp_uid,
             "git_mlcube_url": "mock_url",
-            "mlcube_hash": "",
+            "mlcube_hash": mlcube_yaml_hash,
             "parameters_hash": "",
```
VukW, line 58: Are we checking only mlcube.yaml hash? What about parameters?

hasan7n: @VukW Only `mlcube.yaml`.

hasan7n: It's because `git_mlcube_url` is required, and we set it to `mock_url`. Others are optional, so we don't provide URLs, and hence no hashes are required/checked.

VukW (approved):
LGTM

PR #574, opened 2024-04-05
Disable annotations & automatic consumption of reviewed cases

aristizabal95:
This PR disables the possibility of making annotations through the monitoring tool, to avoid people encountering the bug with annotations. It also disables extracting finalized cases from tarballs to avoid integrating bugged annotations.


hasan7n:
@aristizabal95 This is looking good, nice work! I will test it and let you know.
I may not be aware of all corners of the monitoring tool code, but everything looks great! Just one small thing that I want to double check with you:

https://github.com/aristizabal95/medperf-2/blob/52957db89f7dbf7a4edac38e3e3c6408c69a9678/scripts/monitor/rano_monitor/handlers/reviewed_handler.py#L75-L92

In the code above, I see `on_modified` is handled correctly. However, the `__init__` would still run anyway right? and it looks to me it is "moving assets". Is that expected?

aristizabal95:
Good catch! I'll fix that tomorrow

scripts/monitor/rano_monitor/widgets/subject_details.py
```diff
@@ -130,20 +138,23 @@ def update_subject(self):
...
-            reviewed_button.label = "Mark as finalized"
-            reviewed_button.disabled = False
-        if self.__can_review_brain():
-            brainmask_button.label = "Review brain mask"
-            brainmask_button.disabled = False
+        if ANNOTATIONS_ENABLED:
+            annotations_disabled_msg.display = "none"
+            if is_editor_installed():
+                review_msg.display = "none"
+                review_button.disabled = False
```
VukW, line 151: So you don't make this check if annotations are disabled, right? thus showing `review_msg`?
And also it looks like all the buttons cannot be disabled if not ANNOTATIONS_ENABLED?

VukW: *fixing myself. Buttons ARE disabled by default and we cannot enable them back. So everything is OK

VukW: *fix to fix:) everything is ok with button disabling, but the question about `review_msg`  still stands

aristizabal95:
closing as it is no longer relevant

PR #575, opened 2024-04-09
Improved integration tests

VukW:
- now config.yaml location can be defined with custom value (env var)
- integration tests use a new /tmp/ folder storage with every run
 - After every command db is backed up to /tmp/%dir%/snapshots/ for easy debug
 - Every command is printed before execution
 
 NB: despite of it simplifies tests rerun, one still need to reset db storage manually before tests run:  `./reset_db.sh && python seed.py`


cli/cli_chestxray_tutorial_test.sh
```diff
@@ -13,42 +13,57 @@ wget -P $DIRECTORY https://storage.googleapis.com/medperf-storage/chestxray_tuto
...
+mkdir -p ~/.medperf
+ev mv -f $MEDPERF_STORAGE/.medperf/* ~/.medperf/
+##########################################################
+
+
 echo "=========================================="
 echo "Creating test profiles for each user"
 echo "=========================================="
-medperf profile activate local
+ev medperf profile activate local
```
VukW, line 34: `ev` is a custom function defined in tests_setup: it echoes command before execution

cli/cli_chestxray_tutorial_test.sh
```diff
@@ -13,42 +13,57 @@ wget -P $DIRECTORY https://storage.googleapis.com/medperf-storage/chestxray_tuto
...
+echo "====================================="
+echo "Changing storage to tmp location"
+echo "====================================="
+# this 'move' command is used only for config updates.
+ev medperf storage move -t $MEDPERF_STORAGE
+checkFailed "moving storage failed"
+echo "We have to return the content back manually, so new tmp folder is empty"
+echo "and all the existing data is kept on the old place."
+mkdir -p ~/.medperf
+ev mv -f $MEDPERF_STORAGE/.medperf/* ~/.medperf/
```
VukW, line 27: a crutch: without it
(1) new tmp storage is spoiled with existing data
(2) it might be deleted after test run, so deleting any files you might want to keep

cli/tests_setup.sh
```diff
@@ -14,40 +14,63 @@ SERVER_URL="${SERVER_URL:-https://localhost:8000}"
...
   echo "Cleaning up medperf tmp files"
   echo "====================================="
   # move back storage
   rm -fr $DIRECTORY
-  rm -fr $MEDPERF_STORAGE/**/$SERVER_STORAGE_ID
-  # errors of the commands below are ignored
-  medperf profile activate default
-  medperf profile delete testbenchmark
-  medperf profile delete testmodel
-  medperf profile delete testdata
```
VukW, line 41: no need to fix list of profiles as whole test root would be deleted

cli/cli_chestxray_tutorial_test.sh
aristizabal95: If the intention is to keep test files in case of debugging or similar, I'd prefer to instead specify the location to store them as an argument instead of merging it with the default medperf storage. This is because there's a chance the test files would collide with other files someone might be working on (e.g. I have to do a demonstration of medperf and I'm using the local profile and server for that).

Alternatively, have a flag that disables deleting the tmp folder, and not letting tests run if that folder isn't empty. 

I'd personally prefer tests artifacts to be isolated from the real medperf storage/server if possible.

cli/cli_chestxray_tutorial_test.sh
aristizabal95: Could we have a more descriptive name for this? Or an easy way to identify where this function is defined and what is it used for? I can see myself getting confused by this in the future. Maybe `eval` or `print_eval`. Might be longer but I'd prefer readability.

cli/medperf/config.py
```diff
@@ -44,7 +45,7 @@
 # Storage config
 config_storage = Path.home().resolve() / ".medperf_config"
 logs_storage = Path.home().resolve() / ".medperf_logs"
-config_path = str(config_storage / "config.yaml")
+config_path = getenv("MEDPERF_CONFIG_PATH", str(config_storage / "config.yaml"))
```
aristizabal95, line 48: Defining this env var is now required for all medperf installations?

cli/tests_setup.sh
```diff
@@ -14,40 +14,54 @@ SERVER_URL="${SERVER_URL:-https://localhost:8000}"
...
 MOCK_TOKENS_FILE="$(dirname $(dirname $(realpath "$0")))/mock_tokens/tokens.json"
-
+SQLITE3_FILE="$(dirname $(dirname $(realpath "$0")))/server/db.sqlite3"
 echo "Server URL: $SERVER_URL"
-echo "Storage location: $MEDPERF_SUBSTORAGE"
 
+ev() {
+    local timestamp=$(date +%m%d%H%M%S)
+    local formatted_cmd=$(echo "$@" | sed 's/[^a-zA-Z0-9]\+/_/g' | cut -c 1-50)
+    LAST_SNAPSHOT_PATH="$SNAPSHOTS_FOLDER/${timestamp}_${formatted_cmd}.sqlite3"
```
aristizabal95, line 36: Could we end up clutteraing this folder indefinitely if we're not cleaning it consistently? Would there be a way of defining a max amount of snapshots that can exist at a given time?

aristizabal95 (commented):
Hey Slava, this is looking great! 

I just left a few suggestions and questions. Other than that I think it's looking good!

cli/cli_chestxray_tutorial_test.sh
aristizabal95: Disregard my message above, I misunderstood this section given the snippet of code commented and the accompanying message. This is only executed when moving the storage failed for some reason right? And it's just to restore the moved files to their original locations. Correct?

cli/medperf/config.py
VukW: No, if env is not defined, default value is used (`str(config_storage / "config.yaml")`). `getenv` works similar to `dict.get()` method

cli/tests_setup.sh
VukW: Not exactly
```
tmp/
|-- medperf_tests_20240409140527/
|--|-- snapshots/ # here all the snapshots for that run are stored
|--|--|--....
|--|-- storage/
|--|--|-- .medperf/
|--|--|--|-- ....
|--|-- config.yaml
```

So, this snapshots folder belongs to a specific tests run, not mixing with snapshots from other runs. No need to manage especially snapshots. If you don't need this test results anymore (or you use `-c`/`CLEAN` flag), you delete a whole `medperf_tests_XXXYYY`.

However by default `-c` is set up to `False`. So, we are cluttering `/tmp/` folder with a bunch of `medperf_tests_XXXYYY`.

What do you think guys if we ultimately remove test root folder if tests were passed successfully? And oldish crap after failed tests we can remove manually.

Or I can implement a

cli/cli_chestxray_tutorial_test.sh
VukW: No:)
`medperf storage move` makes two essential actions:
1. updates all paths in config to point to a new folder
2. actually moves the content of `~/.medperf/` to `/tmp/medperf_tests_XXXYYY/storage/.medperf/`

While (1) is what we want for tests (remember, we have a separate temporary config file while running tests), (2) is eww. Because really we want to work with a new clean empty storage folder while running tests (and, and, in particular, to know it's can be safely deleted at the end).

So, this crutch reverts (2) and moves storage content back to `~/.medperf/`.

cli/cli_chestxray_tutorial_test.sh
VukW: > Could we have a more descriptive name for this? Maybe `eval` or `print_eval`. Might be longer but I'd prefer readability.

Sure:) IMO `ev` is more readable though. I pushed the fix (renamed `ev` to `print_eval`). Plz look and let me know - if it is better for you, let's keep it longish.

hasan7n (dismissed):
@VukW man, that's great! this makes lives a lot easier :)
one small suggestion is to move the `Changing storage to tmp location` bash code into setup since it is common, but that's not necessary!
Thanks for this!

VukW:
Hi @hasan7n thanks, you are right, moved config&storage init to `tests_setup.sh`.

Also updated "storage moving": instead of real moving files back and forth (that may be error-prone if broken on the mid) we explicitly point config pathes to a newly created empty tmp folder. That is safer IMO

VukW:
@hasan7n fixed, ready for re-review:)

hasan7n (approved):
Looks great man!

PR #576, opened 2024-04-10
Package without editable flag

VukW:
An ancestor of https://github.com/mlcommons/medperf/pull/550 (branch was renamed)
====
The goal is to make medperf client installable from pypi. This PR is the first part that allows installing package not from repository

A few changes were implemented:

- git branch check updates: check skipped if no repo found.
- During deployment, local server places `cert.crt`, `cert.key` and `mock_tokens/tokens.json` to `~/.medperf_config/.local_server/` so detached medperf client can use it. NB: this change breaks the behavior for existing developers (as path to local server cert is already defined in user config). You should update local config's `certificate` path to `~/.medperf_config/.local_server/cert.crt`.
- `pip install -e ./cli` replaced with `pip install ./cli` everywhere


aristizabal95 (dismissed):
Looks good to me!

hasan7n (approved):
@VukW looks good, although I believe there is no need to move around tokens and certs since we decided that developers will anyway clone the repo  and install it from source, so the added complexity of moving mock tokens file and having a ~ medperf dev folder could be really omitted; nothing is gained I guess?
anyway please feel free to merge or edit

hasan7n:
@VukW ah I guess the auth test needs to have `--cert ..` removed from calling `seed`

VukW:
> @VukW looks good, although I believe there is no need to move around tokens and certs since we decided that developers will anyway clone the repo and install it from source, so the added complexity of moving mock tokens file and having a ~ medperf dev folder could be really omitted; nothing is gained I guess? anyway please feel free to merge or edit

Looked one more time to it.
So we have 2 issues (location of local cert + location of local tokens).

## local tokens

Local tokens path is hardcoded in `config.py`. We can:
1. move it to `config.yaml` to make it configurable (but we either have to  declare it in every profile including `default`, or make a architectural difference between `default` profile and `local`); + every time you recreate config you'd have to remember you need to fix this field manually
2. require path to be passed via env variable (but then you'd have to provide the env var for all your shells - say, customize your `~/.bashrc` or zsh/fish profile) (AFAIR your proposed this, right?)
3. keep as implemented now, cluttering home folder with ~/.medperf_dev/`

IMO it's better to do (2) or (3), as (1) looks more error-prone

## Local cert
It's already defined in `config.yaml`. So two solutions are
1. fill this field manually every time you recreate config
2. keep as implemented now, cluttering home folder with `~/.medperf_dev`

To tell truth I'm confused and not sure which solution I'd prefer as all of them seem to have some issues. But, maybe, it's easier to me to agree with cluttering home folder rather then fix config fields manually after any config reset (??? tho I believe it would not be really often).

Need your opinion, man @hasan7n :)

hasan7n:
closing as outdated

PR #577, opened 2024-04-10
Fix RANO Annotation procedure

aristizabal95:
This PR fixes the underlying issue with the annotation procedure on the RANO monitoring tool. It provides the correct path to use for modifying the annotations. Additionally, it simplifies the procedure for loading annotations from a tarball file, by using a button instead of a watchdog.

Lastly, scripts are included to cover scenarios for users that are affected by the bug


cli/medperf/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.1.0"
+__version__ = "0.1.2"
```
VukW, line 1: 🔥 🔥 🔥 
finally we have a new version:))

scripts/monitor/rano_monitor/tarball_browser.py
```diff
@@ -50,10 +50,12 @@ def set_vars(self, contents_path):
 
     def __get_subjects(self):
         subjects = os.listdir(self.contents_path)
+        subjects = [subject for subject in subjects if not subject.startswith(".")]
         subject_timepoint_list = []
         for subject in subjects:
             subject_path = os.path.join(self.contents_path, subject)
             timepoints = os.listdir(subject_path)
+            timepoints = [timepoint for timepoint in timepoints if not timepoint.startswith(".")]
             subject_timepoint_list += [(subject, tp) for tp in timepoints]
```
hasan7n, line 60: So this is added to avoid the generated MacOS hidden files? And we are assuming no subject names/timepoint would start with a .?

scripts/monitor/rano_monitor/widgets/summary.py
```diff
@@ -80,5 +79,15 @@ def update_summary(self):
 
     def on_button_pressed(self, event: Button.Pressed) -> None:
         event.stop()
-        package_review_cases(self.report, self.dset_path)
-        self.notify(f"{REVIEW_FILENAME} was created on the working directory")
+        pkg_btn = self.query_one("#package-btn", Button)
+        unpkg_btn = self.query_one("#unpackage-btn", Button)
```
hasan7n, line 83: just double-checking id #package-btn should be #package-btns. (Just hunting for typos :))

hasan7n (approved):
@aristizabal95 thanks! I left some comments but most probably they are nothing important.
BTW haven't reviewed the two scripts since I don't think they are relevant.

scripts/monitor/rano_monitor/tarball_browser.py
aristizabal95: Correct! I could make it more specific if we want to handle hidden cases, but I think that would be a very specific edge case to encounter

scripts/monitor/rano_monitor/widgets/summary.py
aristizabal95: Thanks for the check! In this case it shouldn't be, as that would be the container of both buttons, and not the specific button

scripts/monitor/rano_monitor/handlers/reviewed_handler.py
```diff
@@ -1,115 +0,0 @@
...
-            "DataForQC",
-            id,
-            tp,
-            "TumorMasksForQC",
-            "finalized",
-        )
-        if not os.path.exists(dest_path):
-            # Don't try to add reviewed file if the dest path
-            # doesn't exist
-            continue
```
hasan7n, line 54: sorry I thought I already made a comment. just want to point out that this block of code is removed when you moved this logic to `utils`

aristizabal95: Not sure why I decided to remove that, but I'll add it back to all the identified cases. Thanks for checking that!

VukW (approved):
LGTM

PR #579, opened 2024-04-17
Implement Preparation progress dashboard utility script

aristizabal95:
This PR includes the necessary changes for running a preparation dashboard for their mlcube/experiment. Although at the moment is somewhat targeted to RANO, small tweaks may be done to make it generalizable.

This PR includes two things:
- [x] Allow data preparation mlcube owners to retrieve user information for users that are using any of their mlcubes.
- [x] Provide a utility script for visualizing progress on an experiment dataset preparation.


server/user/tests/test_pk.py
```diff
@@ -42,6 +56,7 @@ def setUp(self):
     @parameterized.expand(
         [
             ("user2", status.HTTP_403_FORBIDDEN),
+            ("user3", status.HTTP_200_OK),
             (None, status.HTTP_401_UNAUTHORIZED),
         ]
     )
```
VukW, line 62: Help me plz to read the test / functionality properly. As far as I understand from the code, the following happens:
1. we log in under the `user3`
2. we request some `user1` profile info (`url` is set up with `user1_id`)
3. we expect that `user3` have access to `user1` profile info
Right?

And the test works as follows:
1. `user3` (benchmark owner in reality?) creates an mlcube (data prep mlcube?)
2. `user1` creates dataset that includes aforementioned  data prep mlcube
3. So, benchmark owner has an access to profiles of users - connected datasets' owners

If yes, let me ask a dummy question. May dataset be connected to any other mlcubes different from dataprep, say, to some model mlcube? In this case would (and should) model mlcube owner have an access to dataset's owner profile?

server/user/permissions.py
```diff
@@ -15,3 +16,18 @@ def has_permission(self, request, view):
...
+        if not pk:
+            return False
+        if not request.user.id:
+            return False
+
+        user_datasets_using_owned_mlcube = Dataset.objects.filter(
+            owner=pk, data_preparation_mlcube__owner=request.user
+        )
+
+        return len(user_datasets_using_owned_mlcube) > 0
```
VukW, line 33: 👍

server/user/tests/test_pk.py
aristizabal95: Correct, although in this case we don't assume `user3` to be a benchmark owner necessarily, but a data prep mlcube owner. There's probably going to be a good intersection between those two roles, but not necessarily always.

Regarding your question with model mlcube, it's a good one, and one that will require discussions with the team. The reason this change had to happen here is because we need the users emails to be able to display that information in the dashboard and know who to contact if something is failing

scripts/dashboard/utils.py
```diff
@@ -0,0 +1,30 @@
+import re
+import os
+
+from medperf import config
+
+
+def stage_id2name(stage_str, stages_df):
+    _, code = stage_str.split()
+    code = float(code)
```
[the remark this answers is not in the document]
aristizabal95: Not sure if there's going to be an easy way out of this, and it comes from a failure of thinking about this when designing the stages. Stages code can be an int or a float as defined in the stages.csv file, while reports we get back from the server can also be int or float depending on the stages encountered by the user. That means that for example we have stage `'8'` but the report says it is `stage 8.0` in which case string matching will not work, or would require at some point float casting to harmonize both strings, which would introduce the aforementioned bugs.

It actually appears that pandas can handle float indices for the most part, behaving as expected in most situations. https://pandas.pydata.org/pandas-docs/version/0.13.1/indexing.html#float64index

While I understand the issue with floats, and I'm dissapointed I didn't consider this when I started designing the whole pipeline, given the scenario we're dealing with here, with well defined and predefined floats instead of floats coming out of computations, I think we can work with floating indices.

Do let me know if you have a thought on how to harmonize the strings from the stages and the report so that we can avoid floating points altogether. We could make changes to how reports are generated, but we would still need to have backwards compatibility with those reports that were generated before the change.

hasan7n:
@aristizabal95 Great work man! Nice and clean.

What are `participant-dashboard` and `rano-dashboard` that somehow look very similar? Also which file is the main entrypoint of the dashboard? Maybe a small README file in this folder could be helpful.

Some additional notes:
- We should have a message during dataset submission notifying the user (in a warning color maybe) that their user information will be visible to the owner of the preparation mlcube they specified (we should explicitly mention what will be visible)
- The `initialize` function imported into the dashboard scripts is best to be run at the very very first before importing any other thing, just to make sure that if something breaks because of config initialization, the culprit would be `initialize` itself and not the new scripts. This could be accomplished by perhaps doing something similar to the `__main__.py` file we have in the main medperf code.
- in the permissions code, `len(queryset)>0` is usually better to be replaced by `queryset.exists()`, although in our small-scale deployments and in our specific use-case the efficiency gain will be negligible.

server/user/tests/test_pk.py
VukW: Hm. If there may be a difference between benchmark owner and data prep cube owner, shouldn't we give an access to the BO instead of data prep owner?

scripts/dashboard/utils.py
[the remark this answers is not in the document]
VukW: And we cannot fix reports easily, right? I mean, what would happen if starting from tomorrow report would send stage code as str instead of float?

I understand such a fix is out of scope for this PR, so I'm asking mostly for getting if we can improve this in future

scripts/dashboard/preparation-dashboard.py
```diff
@@ -0,0 +1,318 @@
...
+    all_stages_list = stages_df["status_name"].values.tolist()
+    all_stages = set(all_stages_list)
+    table_cols = set(latest_table.columns)
+    reported_stages = all_stages.intersection(table_cols)
+    sorted_reported_stages = [
+        stage for stage in all_stages_list if stage in reported_stages
+    ]
+
+    stages = latest_table[sorted_reported_stages]
+    stages["Unknown"] = 1 * (stages.sum(axis=1) == 0)
```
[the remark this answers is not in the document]
VukW: Ah, I got. bool2int conversion

aristizabal95:
@hasan7n Thanks! I had forgotten to push the latest commit that removed the `rano-dashboard` script. I basically tried to generalize the script so that it no longer was rano-specific.

I applied your suggestions. Thanks!

cli/medperf/commands/dataset/prepare.py
```diff
@@ -309,8 +309,13 @@ def prompt_for_report_sending_approval(self):
             + " dataset subjects have reached Stage 1, and that 60% of your dataset subjects"
             + " have reached Stage 3:"
         )
+        warning = (
+            "Note that reports will be visible by the Data Preparation MLCube owner and by the"
+            " Benchmark owner to keep track of progress and provide support if needed."
+        )
         config.ui.print(msg)
         dict_pretty_print(example)
+        config.ui.print_warning(warning)
```
hasan7n, line 318: We should have a message during dataset submission notifying the user (in a warning color maybe) that their user information will be visible to the owner of the preparation mlcube they specified (we should explicitly mention what will be visible)

hasan7n: i.e., during "dataset submission", not here. when we ask the user for confirmation of submission, in the same prompt we write a yellow text stating that "your email address will be visible to the data preparation MLCube owner".

scripts/dashboard/medperf_dashboard/get_data.py
```diff
@@ -0,0 +1,126 @@
...
+
+app = Typer()
+
+
+def get_dsets(mlcube_id):
+    dsets = Dataset.all(filters={"mlcube": mlcube_id})
+    dsets = [dset.todict() for dset in dsets]
+    for dset in dsets:
+        user_id = dset["owner"]
+        dset["user"] = config.comms.get_user(user_id)
```
hasan7n, line 20: how many API calls are we expecting here? maybe we can have an endpoint that returns all of this

hasan7n: I expect here a large number of unnecessary API calls (this includes multiple datasets for the same user, so we are talking more than 30 maybe?)
let's have an issue to replace this code later. in the FL PR we will have an endpoint that lists experiment datasets and their owners information so we can reuse the logic defined there to make this better.

scripts/dashboard/medperf_dashboard/assets/stages.csv
```diff
@@ -0,0 +1,51 @@
...
+If using the monitoring tool, create a new tarball with the previously untarred files, as well as finalized cases. You can do so with the following command
+
+``` tar -czf reviewed_cases.tar.gz -C review_cases . ```
+
+Place this new tarball file on the directory the monitoring tool is running (displayed on the monitoring tool header). The tool will automatically detect the finalized cases and place them in the expected locations.
+
+### Brain Mask Correction
+There might be situations where the brain mask is not correct. If working locally you may use the `Review Brain Mask` button to automatically make brain mask corrections with ITK-Snap. If instead you're using the previously menitoned tarball, you may find the brain mask segmentation as `brainMask_fused.nii.gz`. Please make the necessary corrections under that file, overwriting the contents. Once the new tarball is created and placed within the directory the monitoring tool is running, it will be automatically identified and placed in the correct location.
+**NOTE** Modifying the brain mask invalidates the tumor segmentations, and re-runs the tumor segmentation procedure using this new mask.
+",,mediumseagreen
```
hasan7n, line 37: is the text here intentional?

hasan7n:
@aristizabal95 I left few comments.

server/user/tests/test_pk.py
hasan7n: @VukW that's true; for the experiment owner not the benchmark owner since in RANO there is no benchmark. but since until this moment we don't have training experiment logic, we had only prep cube so the current permissions are temporary until all hospitals associate with the training experiment that we will create soon

server/user/tests/test_pk.py
```diff
@@ -42,6 +56,7 @@ def setUp(self):
     @parameterized.expand(
         [
             ("user2", status.HTTP_403_FORBIDDEN),
+            ("user3", status.HTTP_200_OK),
```
VukW, line 59:
```suggestion
            ("user3", status.HTTP_200_OK),
            ("user1", status.HTTP_200_OK),
```

not connected with your PR actually - just adds a test for `IsOwnUser` for better coverage

scripts/dashboard/medperf_dashboard/assets/stages.csv
VukW: It seems to me this is just a multiline str value of comment field for the `-5` stage. At least csv is parsed successfully

scripts/dashboard/medperf_dashboard/get_data.py
```diff
@@ -0,0 +1,126 @@
...
+
+from .utils import get_institution_from_email, get_reports_path, stage_id2name
+
+from typer import Typer, Option, run
+
+app = Typer()
+
+
+def get_dsets(mlcube_id):
+    dsets = Dataset.all(filters={"mlcube": mlcube_id})
```
VukW, line 16: I cannot get quite well how does this work? `Dataset.all` uses `config.comms` Object but it is properly initialized only `medperf/cli/medperf/init.py:initialize` that doesn't seem to be called. I don't have a properly set up env but it looks like `medperf-dashboard` works (I mean, `config.comms` is initialized and really tries to reach the server)

scripts/dashboard/medperf_dashboard/assets/stages.csv
VukW: And I don't see any usage of this `comment` field in the dashboard. I believe it is here just because `stages.csv` file was taken as it is used in monitor (?). Anyway, it does not break anything

scripts/dashboard/medperf_dashboard/assets/stages.csv
```diff
@@ -0,0 +1,51 @@
...
+7,Annotation Confirmed,,,silver
+8,Done,,,green
+-0.101,Generate Report Unhandled Error,,,tomato
+-1.101,Initial Validation Unhandled Error,,,firebrick
+-2.101,Nifti Conversion Unhandled Error,,,teal
+-3.101,Brain Extract Unhandled Error,,,saddlebrown
+-4.101,Tumor Extract Unhandled Error,,,sandybrown
+-5.101,Manual Annotation Unhandled Error,,,peru
+-6.101,Label Segmentation Comparison Unhandled Error,,,moccasin
+-7.101,Annotation Confirmation Unhandled Error,,,mediumvioletred
```
VukW, line 51: @aristizabal95 are you sure you want to upload this file to the main repo? It seems not a generalizeable IMO and relates just to one specific project. But I don't see any issues in comitting it here also

VukW (approved):
LGTM, don't see any blockers for merge.
Thanks for the clean code & well-structured code, man, it was a pleasure to read:)

VukW (approved):

PR #586, opened 2024-04-29
refactor entities

hasan7n:
Refactor entities and use inheritance to not repeat code. This is needed since for FL support we are adding 4 new entities with similar interface.

This PR also removes the complexity and notion of local-only for listing entities, and defines an `unregistered` notion

One downside for this change is intellisense in code editors is now broken. To solve this (in another PR), we can use type hints when retrieving entities, and refactor the inheritance of schemas.

### DIFF
--- a/cli/medperf/commands/dataset/prepare.py
+++ b/cli/medperf/commands/dataset/prepare.py
@@ -165,8 +165,7 @@ class DataPreparation:
 
     def run_prepare(self):
         report_sender = ReportSender(self)
-        if self.allow_sending_reports:
-            report_sender.start()
+        report_sender.start()
 
         prepare_params = {
             "data_path": self.raw_data_path,
@@ -190,18 +189,15 @@ class DataPreparation:
                 )
         except Exception as e:
             # Inform the server that a failure occured
-            if self.allow_sending_reports:
-                report_sender.stop("failed")
+            report_sender.stop("failed")
             raise e
-        except KeyboardInterrupt:
+        except KeyboardInterrupt as e:
             # Inform the server that the process is interrupted
-            if self.allow_sending_reports:
-                report_sender.stop("interrupted")
-            raise
+            report_sender.stop("interrupted")
+            raise e
 
         self.ui.print("> Cube execution complete")
-        if self.allow_sending_reports:
-            report_sender.stop("finished")
+        report_sender.stop("finished")
 
     def run_sanity_check(self):
         sanity_check_timeout = config.sanity_check_timeout
@@ -313,8 +309,12 @@ class DataPreparation:
         dict_pretty_print(example)
 
         msg = (
-            " \nDo you approve the automatic submission of summaries similar to the one above"
-            + " to the MedPerf Server throughout the preparation process?[Y/n]"
+            "\nYou can decide whether or not to send information about your dataset preparation"
+            + "\nProgress. Keep in mind that information about the execution status of the pipeline"
+            + "\nwill be sent regardless (whether the pipeline is running, finished or failed)"
+            + "\nto identify issues with the preparation procedure. Do you approve the automatic"
+            + "\nsubmission of summaries similar to the one above to the MedPerf Server throughout"
+            + "\nthe preparation process?[Y/n]"
         )
 
         self.allow_sending_reports = approval_prompt(msg)
@@ -327,7 +327,12 @@ class DataPreparation:
             return self._send_report(report_metadata)
 
     def _send_report(self, report_metadata):
-        report_status_dict = self.__generate_report_dict()
+        if self.dataset.for_test:
+            # Test datasets don't have a registration on the server
+            return
+        report_status_dict = {}
+        if self.allow_sending_reports:
+            report_status_dict = self.__generate_report_dict()
         report = {"progress": report_status_dict, **report_metadata}
         if report == self.dataset.report:
             # Watchdog may trigger an event even if contents didn't change
--- a/cli/medperf/tests/commands/dataset/test_prepare.py
+++ b/cli/medperf/tests/commands/dataset/test_prepare.py
@@ -8,7 +8,7 @@ import pytest
 
 from medperf.tests.mocks.dataset import TestDataset
 from medperf.tests.mocks.cube import TestCube
-from medperf.commands.dataset.prepare import DataPreparation
+from medperf.commands.dataset.prepare import DataPreparation, Observer
 
 PATCH_REGISTER = "medperf.commands.dataset.prepare.{}"
 
@@ -54,37 +54,58 @@ def test_get_prep_cube_downloads_cube_file(mocker, data_preparation, cube):
     spy.assert_called_once()
 
 
+@pytest.mark.parametrize("dataset_for_test", [False, True])
+def test_prepare_with_test_data_doesnt_send_reports(
+        mocker, data_preparation, dataset_for_test, cube, comms, fs
+):
+    # Arrange
+    data_preparation.dataset.for_test = dataset_for_test
+    mocker.patch.object(cube, "run")
+    mocker.patch.object(data_preparation.dataset, "write")
+    mocked_obs = mocker.create_autospec(spec=Observer)
+    mocker.patch(PATCH_REGISTER.format("Observer", side_effect=mocked_obs))
+    send_report_spy = mocker.patch.object(comms, "update_dataset")
+
+    # Act
+    data_preparation.run_prepare()
+
+    # Assert
+    if dataset_for_test:
+        send_report_spy.assert_not_called()
+    else:
+        send_report_spy.assert_called()
+
+
 @pytest.mark.parametrize("allow_sending_reports", [False, True])
 def test_prepare_runs_then_stops_report_handler(
-    mocker, data_preparation, allow_sending_reports, cube
+    mocker, data_preparation, allow_sending_reports, cube, comms, fs
 ):
     # Arrange
     data_preparation.allow_sending_reports = allow_sending_reports
     mocker.patch.object(cube, "run")
-    start_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.start"))
-    stop_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.stop"))
+    mocker.patch.object(data_preparation.dataset, "write")
+    mocked_obs = mocker.create_autospec(spec=Observer)
+    mocker.patch(PATCH_REGISTER.format("Observer"), side_effect=mocked_obs)
+    gen_report_spy = mocker.patch(PATCH_REGISTER.format("DataPreparation._DataPreparation__generate_report_dict"))
 
     # Act
     data_preparation.run_prepare()
 
     # Assert
     if allow_sending_reports:
-        start_spy.assert_called_once()
-        stop_spy.assert_called_once_with("finished")
+        gen_report_spy.assert_called()
     else:
-        start_spy.assert_not_called()
-        stop_spy.assert_not_called()
+        gen_report_spy.assert_not_called()
 
 
-@pytest.mark.parametrize("allow_sending_reports", [False, True])
 def test_prepare_runs_then_stops_report_handler_on_failure(
-    mocker, data_preparation, allow_sending_reports, cube
+    mocker, data_preparation, cube
 ):
     # Arrange
     def _failure_run(*args, **kwargs):
         raise Exception()
 
-    data_preparation.allow_sending_reports = allow_sending_reports
+    data_preparation.allow_sending_reports = True
     mocker.patch.object(cube, "run", side_effect=_failure_run)
     start_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.start"))
     stop_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.stop"))
@@ -94,23 +115,18 @@ def test_prepare_runs_then_stops_report_handler_on_failure(
         data_preparation.run_prepare()
 
     # Assert
-    if allow_sending_reports:
-        start_spy.assert_called_once()
-        stop_spy.assert_called_once_with("failed")
-    else:
-        start_spy.assert_not_called()
-        stop_spy.assert_not_called()
+    start_spy.assert_called_once()
+    stop_spy.assert_called_once_with("failed")
 
 
-@pytest.mark.parametrize("allow_sending_reports", [False, True])
 def test_prepare_runs_then_stops_report_handler_on_interrupt(
-    mocker, data_preparation, allow_sending_reports, cube
+    mocker, data_preparation, cube
 ):
     # Arrange
     def _failure_run(*args, **kwargs):
         raise KeyboardInterrupt()
 
-    data_preparation.allow_sending_reports = allow_sending_reports
+    data_preparation.allow_sending_reports = True
     mocker.patch.object(cube, "run", side_effect=_failure_run)
     start_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.start"))
     stop_spy = mocker.patch(PATCH_REGISTER.format("ReportSender.stop"))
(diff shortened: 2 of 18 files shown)


cli/cli_tests.sh
```diff
@@ -186,7 +185,7 @@ echo "Running data submission step"
 echo "====================================="
 medperf dataset submit -p $PREP_UID -d $DIRECTORY/dataset_a -l $DIRECTORY/dataset_a --name="dataset_a" --description="mock dataset a" --location="mock location a" -y
 checkFailed "Data submission step failed"
-DSET_A_UID=$(medperf dataset ls | grep dataset_a | tr -s ' ' | cut -d ' ' -f 1)
+DSET_A_UID=$(medperf dataset ls | grep dataset_a | tr -s ' ' | awk '{$1=$1;print}' | cut -d ' ' -f 1)
```
VukW, line 188: Looks good, but why the previous version failed?

hasan7n: previously, unregistered entities were also listed by default (so this will list the demo dataset hash), but currently no (since we now require --unregistered to list unregistered only). unregistered entities have a long hash instead of ID.
due to how we use python's `tabulate`, having a long entry in one of the rows will remove leading spaces of other shorter entries, hence we didn't want the `awk` command to remove leading spaces. but now all entries are short so you may have leading spaces

cli/medperf/entities/dataset.py
```diff
@@ -86,48 +94,8 @@ def is_ready(self):
...
-            dsets = [cls(**meta) for meta in dsets_meta]
-        except CommunicationRetrievalError:
-            msg = "Couldn't retrieve all datasets from the server"
-            logging.warning(msg)
-
-        return dsets
-
     @classmethod
-    def __remote_prefilter(cls, filters: dict) -> callable:
+    def _Entity__remote_prefilter(cls, filters: dict) -> callable:
```
VukW, line 98: This staff looks like a magic command. How is it called? in `interface` I still see `__remote_prefilter` is called
https://github.com/hasan7n/medperf/blob/refactor-entities/cli/medperf/entities/interface.py#L68

hasan7n: yeah it's because we are overriding a superclass method that is private. I am thinking of making it public to avoid this kind of magic/weird things

VukW: Can we make it "usual" private via naming `_remote_prefilter`? AFAIU in that case the method is visible to the children (i.e. can be overridden easily), but is kinda "hidden" from external code (i.e. if external code would call this method, any IDE would highlight that method is protected and you should not refer to it directly)

VukW: Anyway, I agree, if we cannot use usual private naming, it's worth to make it public: right now IDE even does not show to me where the method is overridden (as it happens only dynamically during code execution)

cli/medperf/entities/schemas.py
```diff
@@ -74,7 +74,7 @@ class Config:
         use_enum_values = True
 
 
-class MedperfSchema(MedperfBaseSchema):
+class MedperfSchema(BaseModel):
```
VukW, line 77: Now the code works as `MedperfSchema` is always used together with `Entity`. However, don't it look strange that `MedperfSchema` is not connected with `MedperfBaseSchema` anymore? Maybe we can refactor it a bit more?
As far as I see, the key is `TestReport` that should inherit from `MedperfBaseSchema` but not have fields from `MedperfSchema` (?) I don't see a good solution here at glance, but I believe you thought about it, and all possible solutions of joining everything together have their own key disadvantages, right?

cli/medperf/entities/interface.py
```diff
@@ -1,77 +1,215 @@
...
+    def get_metadata_filename():
+        raise NotImplementedError()
+
+    @staticmethod
+    def get_comms_uploader():
+        raise NotImplementedError()
+
+    @property
+    def identifier(self):
+        return self.id or self.generated_uid
```
VukW, line 33: There is no `id` or `generated_uid` in neither `Entity` or `MedperfBaseSchema`. But there is `id` in `MedperfSchema`.
It seems that `id` should present everywhere so I'd propose to move it from `MedperfSchema` to `Entity` directly;
while `generated_uid` does not exist say for dataset, so it would be better if any entity that uses `generated_uid` overwrite this property on its own

cli/medperf/entities/interface.py
```diff
@@ -1,77 +1,215 @@
...
 
-class Uploadable:
-    @abstractmethod
     def upload(self) -> Dict:
         """Upload the entity-related information to the communication's interface
 
         Returns:
             Dict: Dictionary with the updated entity information
         """
+        if self.for_test:
```
VukW, line 198: `for_test` is never defined in `Entity` or its parents

cli/medperf/entities/interface.py
```diff
@@ -1,77 +1,215 @@
...
+    @staticmethod
+    def get_comms_retriever():
+        raise NotImplementedError()
+
+    @staticmethod
+    def get_metadata_filename():
+        raise NotImplementedError()
+
+    @staticmethod
+    def get_comms_uploader():
```
VukW, line 28:
```suggestion
    def get_comms_uploader() -> `Callable[dict, dict]`:
```

cli/medperf/entities/interface.py
VukW: Hmm. I noticed in old code `Dataset` had both `id` and `generated_uid` while in new code it seems to not have id?..
Also TestResult seems not to have `id` but `generated_uid` only?.. Anyway, it's worth to declare explicitly these fields

cli/medperf/entities/report.py
```diff
@@ -52,71 +59,14 @@ def set_results(self, results):
...
-        report_file = os.path.join(report_path, config.test_report_file)
-        if not os.path.exists(report_file):
-            raise InvalidArgumentError(
-                f"The requested report {local_uid} could not be retrieved"
-            )
-        with open(report_file, "r") as f:
-            report_info = yaml.safe_load(f)
-        return report_info
+    def get(cls, report_uid: str, local_only: bool = False) -> "TestReport":
+        return super().get(report_uid, local_only=True)
```
VukW, line 69:
```suggestion
    def get(cls, uid: str, local_only: bool = False) -> "TestReport":
        """Gets an instance of the TestReport. ignores local_only inherited flag as TestReport is always a local entity.
        Args:
            uid (str): Report Unique Identifier
            local_only (bool): ignored. Left for aligning with parent Entity class
        Returns:
            TestReport: Report Instance associated to the UID
        """
        return super().get(uid, local_only=True)
```
(1) to highlight it's not a bug and `local_only` is ignored conciously,
(2) still keep the same function signature (in case if we once want to make `cls.get(uid=entity_uid)` that would look proper from the IDE side but would fail on TestReport entities)

cli/medperf/entities/cube.py
```diff
@@ -155,36 +104,12 @@ def get(cls, cube_uid: Union[str, int], local_only: bool = False) -> "Cube":
...
-                cube = cls.__remote_get(cube_uid)
-            except CommunicationRetrievalError:
-                logging.warning(f"Getting MLCube {cube_uid} from comms failed")
-                logging.info(f"Retrieving MLCube {cube_uid} from local storage")
-                cube = cls.__local_get(cube_uid)
-
+        cube = super().get(cube_uid, local_only)
         if not cube.is_valid:
             raise InvalidEntityError("The requested MLCube is marked as INVALID.")
         cube.download_config_files()
```
VukW, line 110: if `local_only`, should we still download config files?

cli/medperf/entities/interface.py
```diff
@@ -1,77 +1,215 @@
...
+            uid (str): uid of the local entity
 
         Returns:
-            dict: the display dictionary
+            dict: information of the entity
         """
+        logging.info(f"Retrieving {cls.get_type()} {uid} from local storage")
+        storage_path = cls.get_storage_path()
+        metadata_filename = cls.get_metadata_filename()
+        bmk_file = os.path.join(storage_path, str(uid), metadata_filename)
```
VukW, line 168: Just a small nitpick: I believe `bmk` means `benchmark` while here it's not necessarily a benchmark anymore:)

cli/medperf/entities/interface.py
```diff
@@ -1,77 +1,215 @@
...
+                )
+            return cls.__unregistered_all()
+        return cls.__remote_all(filters=filters)
+
+    @classmethod
+    def __remote_all(cls, filters: dict) -> List["Entity"]:
+        comms_fn = cls.__remote_prefilter(filters)
+        entity_meta = comms_fn()
+        entities = [cls(**meta) for meta in entity_meta]
+        return entities
```
VukW, line 71: so we are not saving fetched entities, right?

VukW: I mean is this conscious? Are there +/- comparing to use `__remote_get` and writing all the fetched entities?

cli/medperf/entities/interface.py
```diff
@@ -1,77 +1,215 @@
...
+        except StopIteration:
+            msg = f"Couldn't iterate over the {cls.get_type()} storage"
+            logging.warning(msg)
+            raise MedperfException(msg)
+
+        for uid in uids:
+            if uid.isdigit():
+                continue
+            meta = cls.__get_local_dict(uid)
+            entity = cls(**meta)
```
VukW, line 88: Might be replaced with `__local_get` call (if you wish)
Though it would print a separate log message for every entity

cli/medperf/commands/list.py
```diff
@@ -10,27 +10,29 @@ class EntityList:
...
+        unregistered: bool = False,
         mine_only: bool = False,
         **kwargs,
     ):
         """Lists all local datasets
 
         Args:
-            local_only (bool, optional): Display all local results. Defaults to False.
+            unregistered (bool, optional): Display only local unregistered results. Defaults to False.
             mine_only (bool, optional): Display all current-user results. Defaults to False.
```
VukW, line 21:
```suggestion
            mine_only (bool, optional): Display all registered current-user results. Defaults to False.
```

cli/medperf/commands/compatibility_test/utils.py
```diff
@@ -138,23 +138,23 @@ def create_test_dataset(
...
 
     # prepare/check dataset
     DataPreparation.run(dataset.generated_uid)
 
     # update dataset generated_uid
-    old_path = dataset.path
-    generated_uid = get_folders_hash([dataset.data_path, dataset.labels_path])
-    dataset.generated_uid = generated_uid
-    dataset.write()
-    if dataset.input_data_hash != dataset.generated_uid:
```
VukW, line 154: It seems previously  old uid was stored in `input_data_hash` but we do not update it. Why?

cli/medperf/commands/compatibility_test/utils.py
```diff
@@ -138,23 +138,23 @@ def create_test_dataset(
     # TODO: existing dataset could make problems
     # make some changes since this is a test dataset
     config.tmp_paths.remove(data_creation.dataset.path)
-    data_creation.dataset.write()
```
VukW, line 141: In oldish code we were writing dataset multiple times while now we write it only at the very end if hash differs (?).
I don't know if it was made consciously or accidentally; It seems to me new code is working properly also, but I'm not 100% sure we didn't break some legacy crutch that was located there. 
So, I'm just highlighting the change: if you don't expect it, can you look plz one more time to assure it doesn't break anything

VukW:
1. I rrreally like how the entities code is simplified. Now it's much easier to be read, thanks a lot man!
2. In the previous code version `TestReport` was a bit more separated from other entities as it doesn't include any remote-entities logic. In new code as it inherits from `Entity` that assumes all ancestors are uploadable, it contain some mock-up methods like `get_comms_uploader` that are filled with `NotImplementedError`. Despite of it may look a bit dirty that we have a `NIErr` functions in the running code, I personally like this solution more than previous one; it seems cleaner to just "close" some unused remote methods rather then separate TestReport from other entities on the architecture & inheritance level. So, :+1: for the new solution
3. I didn't review test code quite well, sorry. I just hope it works. It's a cruel irony you have to spend so much efforts on fixing tests (I'm pretty sure it wasn't easy) and I cannot appreciate it as reading pytest code is too difficult 😅
4. The only major question IMO is to improve `Entity` code even more and declare all required fields / properties there (in constructor?): `id` / `generated_uid` / `for_test`, maybe smth else. The goal is to show IDE that Entity __always__ have these fields and it's safe to refer to them.
5. I also noted your confusion concerning expected method results types, but also didn't get any reasonable solution while reviewing the code, sorry

cli/medperf/entities/dataset.py
hasan7n: if we name it `_remote_prefilter`, it will not really override the superclass's method because it's private (tested it then figured out that it should be named in this weird way so that it works)

So yeah, I am gonna make it public for ease of understanding the code and for better editors intellisense

cli/medperf/entities/schemas.py
hasan7n: you are right, it now looks strange. I think I have an idea, will push a change soon

cli/medperf/entities/dataset.py
VukW: I mean, call it `__remote_prefilter` + `@staticmethod` instead of `@classmethod`. We don't work with classes here really, and static methods are simpler to override

cli/medperf/entities/interface.py
hasan7n: we were using `generated_uid` in two different contexts, which makes it confusing:
The dataset object has in its Server schema a `generated_uid`, which actually should be really renamed to `output_hash` but such renaming may cause us pain now.
All entities have some sort of "local ID", which we were refering to by `generated_uid`.
Now, I pushed a change that removes such an ambiguity by renaming the general `generated_uid` to `local_id`. Dataset's `local_id` happens to be the `generated_uid`, but I hope that this is not confusing anymore (may be confusing to people who already read the code before this change, but I think it's clearer for people reading the code for the first time).
and also I defined a `local_id` in the superclass.

cli/medperf/entities/interface.py
hasan7n: One thing we had in mind is that it could be an overkill to save them all locally (imagine a benchmark having 100 results for example). I don't see benefit of writing them btw.

cli/medperf/commands/compatibility_test/utils.py
hasan7n: `input_data_hash` is meant to be the hash of the raw data, which we know it doesn't change after preparation in this code snippet.
`generated_uid` (aka `output_data_hash`) is set to `input_data_hash` before preparation just because "technically the data is not prepared so rawdata == outputdata", but that is not really critical

hasan7n:
@VukW 
Thanks for the great comments. One major thing I pushed is that I went further and made the `TestReport` have the common attributes of other entities (id, created_at, owner, ....) so that we get rid of `MedPerfbaseschema` vs `medperfschema`, hence we don't make the inheritence complicated...

Regarding expected methods results type and intellisense, it seems to work when we do something like `dataset: Dataset = Dataset.get(...)`. It's an overkill to make such a change in all the code so I think while we develop other stuff in other PRs we can start using type hints...

cli/medperf/entities/benchmark.py
```diff
@@ -58,10 +58,12 @@ def __init__(self, *args, **kwargs):
         """
         super().__init__(*args, **kwargs)
 
-        self.generated_uid = f"p{self.data_preparation_mlcube}m{self.reference_model_mlcube}e{self.data_evaluator_mlcube}"
+    @property
+    def local_id(self):
+        return self.name
```
VukW, line 63: AFAIU now generated_uid is replaced with local_id. But where is this field `name` taken from? I don't see how it is built now...

hasan7n: now the local ID is changed from `p{self.data_preparation_mlcube}m{self.reference_model_mlcube}e{self.data_evaluator_mlcube}` to just the name of the benchmark. That's it

VukW (dismissed):
LGTM

VukW (dismissed):

VukW (approved):

VukW:
Auth test still fails because of `webdriver-manager` requires Chrome == 125 (by default), while new version installs Chrome == 127(?)
https://github.com/mlcommons/medperf/actions/runs/9464826888/job/26072984898?pr=586
Can we solve this in a separate PR and merge this one?

PR #600, opened 2024-06-20
Specify numpy version

aristizabal95:
This PR fixes an issue between Pandas and Numpy, by installed a version of numpy that is known to work with the version of pandas being used.


VukW (approved):
Lgtm

PR #601, opened 2024-06-24
Improve Monitoring Tool UX related to annotations

aristizabal95:
This PR adds a few QOL improvements to the annotation procedure. It does so by:
- Displaying when a subject contains finalized annotations
- Displaying when a subject's brain mask has been modified
- Displaying a hint on the summary indicating the user they have to run the pipeline to proceed after annotations

These have been points of confusion for several users, thinking their changes have not been taken into account because the monitoring tool doesn't update until the pipeline is running. This PR should improve that experience.

### DIFF
--- a/scripts/monitor/rano_monitor/utils.py
+++ b/scripts/monitor/rano_monitor/utils.py
@@ -388,6 +388,7 @@ def unpackage_reviews(file, app, dset_data_path):
                 delete(target_file, dset_data_path)
             tar.extract(member, dest)
 
+
 def brain_has_been_reviewed(brainpath, backup_brainpath):
     if not os.path.exists(backup_brainpath):
         return False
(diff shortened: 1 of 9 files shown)


scripts/monitor/rano_monitor/utils.py
```diff
@@ -387,3 +387,58 @@ def unpackage_reviews(file, app, dset_data_path):
...
             tar.extract(member, dest)
+
+
+def brain_has_been_reviewed(brainpath, backup_brainpath):
+    if not os.path.exists(backup_brainpath):
+        return False
+
+    brain_hash = get_hash(brainpath)
+    backup_hash = get_hash(backup_brainpath)
+    return brain_hash != backup_hash
```
VukW, line 398: To tell truth I don't understand the whole pipeline logics quite well. Can you explain plz why difference between brain_hash and backup_hash means anything was reviewed? For me it means only that the content was modified comparing to the backup...

aristizabal95: The brain mask will only change if someone opened the file and changed the contents, so this is why this check works to identify if the brain mask has been reviewed. We don't care about brain masks that have been opened but not modified.

As for tumor segmentations, there are situations in which someone might review the file but leave it unchanged, and we do need to identify those cases to be able to move forward. That's why for tumor segmentations we instead check the existence of that segmentation in a "finalized" folder. That file will only be there if a user puts it there (either manually or by using the monitoring tool "set as finalized" button).

aristizabal95:
Closing for a bigger PR with more UX improvements

PR #602, opened 2024-07-03
Monitor UX Improvements

aristizabal95:
This PR provides many QoL improvements to the monitoring tool, addressing many points of confusion and annoyance encountered by most users.

This PR includes:

- Packaging and loading review cases running on a separate thread, to avoid blocking the UI.
- Displaying whether a case has been finalized or has a modified brain mask directly in the subject list
- Displaying hints in the summary page as to how to proceed after annotations
- Displaying the number of cases under a given stage instead of just the percentage
- Adding a search bar to filter out the list of subjects by subject identifier or stage

Screenshot 2024-07-03 at 12 31 04 PM


scripts/monitor/rano_monitor/widgets/summary.py
```diff
@@ -61,10 +68,14 @@ def update_summary(self):
             status_percents["DONE"] = 0.0
 
         package_btns.display = "MANUAL_REVIEW_REQUIRED" in status_percents
+        hint_msg.display = package_btns.display
```
VukW, line 71: So hint is shown only if there are items that require manual review. What if not? Why not to show hint if there are some items done, and a few say are extra modalities?

scripts/monitor/rano_monitor/widgets/tarball_subject_view.py
```diff
@@ -55,9 +49,15 @@ def on_mount(self):
     def update_status(self):
         tumor_status = self.query_one(".tumor-status", Static)
         brain_status = self.query_one(".brain-status", Static)
-        if self.__tumor_has_been_finalized():
+
+        id, tp = self.subject.split("|")
+        finalized_tumor_path = os.path.join(self.contents_path, id, tp, "finalized")
```
VukW, line 54: This finalized tumor path differs from what is building at `get_finalized_tumor_path`, is it expected?

VukW:
Can't tell unluckily I understood all the staff around UI, so hope you did 😅

scripts/monitor/rano_monitor/widgets/tarball_subject_view.py
aristizabal95: This is expected, because the tarball version of the monitoring tool doesn't interact with the same paths, but the ones related to the tarball file

scripts/monitor/rano_monitor/widgets/summary.py
aristizabal95: That's as good point. I can remove the conditional and keep it shown all the time

PR #603, opened 2024-07-09
Allow passing itksnap path/command as an argument for the monitoring tool

aristizabal95:
In a few cases, people haven't been able to run the monitoring tool because they don't have itksnap cli installed as the monitoring tool expects, and they have valid reasons for this (multiple versions of itksnap, not having permissions to move itksnap to the binary folder, etc).

This PR allows users to pass an optional argument when running the monitoring tool to define where itksnap is found.

### DIFF
--- a/scripts/monitor/rano_monitor/utils.py
+++ b/scripts/monitor/rano_monitor/utils.py
@@ -9,7 +9,6 @@ from subprocess import DEVNULL, Popen
 import pandas as pd
 import yaml
 from rano_monitor.constants import (
-    REVIEW_COMMAND,
     BRAINMASK_BAK,
     DEFAULT_SEGMENTATION,
     BRAINMASK,
--- a/scripts/monitor/rano_monitor/utils.py
+++ b/scripts/monitor/rano_monitor/utils.py
@@ -34,7 +34,7 @@ def get_hash(filepath: str):
     return file_hash
 
 
-def run_editor(t1c, flair, t2, t1, seg, label, cmd=REVIEW_COMMAND):
+def run_editor(t1c, flair, t2, t1, seg, label, cmd):
     review_cmd = "{cmd} -g {t1c} -o {flair} {t2} {t1} -s {seg} -l {label}"
     review_cmd = review_cmd.format(
         cmd=cmd,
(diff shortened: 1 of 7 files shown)


scripts/monitor/rano_monitor/utils.py
```diff
@@ -37,7 +37,7 @@ def get_hash(filepath: str):
 def run_editor(t1c, flair, t2, t1, seg, label, cmd=REVIEW_COMMAND):
```
VukW, line 37: If now command path is passing explicitly, do you need to keep a default value for `cmd` here? it may be error-prone if you don't fix a call somewhere

scripts/monitor/rano_monitor/utils.py
```diff
@@ -48,7 +48,9 @@ def run_editor(t1c, flair, t2, t1, seg, label, cmd=REVIEW_COMMAND):
     Popen(review_cmd.split(), shell=False, stdout=DEVNULL, stderr=DEVNULL)
 
 
-def review_tumor(subject: str, data_path: str, labels_path: str):
+def review_tumor(
+    subject: str, data_path: str, labels_path: str, review_cmd: str = REVIEW_COMMAND
+):
```
VukW, line 53: Same - do you need default value here?

scripts/monitor/rano_monitor/utils.py
```diff
@@ -65,10 +67,18 @@ def review_tumor(subject: str, data_path: str, labels_path: str):
...
+        t2w_file,
+        t1n_file,
+        under_review_file,
+        label_file,
+        cmd=review_cmd,
+    )
 
 
-def review_brain(subject, labels_path, data_path=None):
+def review_brain(subject, labels_path, data_path=None, review_cmd=REVIEW_COMMAND):
```
VukW, line 81: Same

VukW (dismissed):

scripts/monitor/rano_monitor/utils.py
aristizabal95: Completely agree. I'm removing the default value

VukW (dismissed):
LGTM (after liter fix)

VukW (approved):

PR #615, opened 2024-09-10
[WIP] Config refactoring

VukW:
Right now configuration process is quite a complex: settings are spread over two places (`config.py` and yaml profiles), both are rewritten multiple times during app init, with possible overriding by CLI params. At the same time some settings like `UI`, `Auth`, `Comms` objects are fixed just once at the very start of the app and cannot be reconfigured easily later.
However, for web-UI we aim to support a runtime change of profile / auth, so user can log in and use web-ui without restarting CLI command. For that I do need to refactor config init approach - and, at the same time, it's possible to simplify configuration init for the future usage.


hasan7n:
closing as outdated

PR #606, opened 2024-07-23
Web UI - stage1, readonly part

VukW:
Adds a local web UI to use instead of CLI UI.

On this stage, only read-only views are implemented:
- list of benchmarks
- list of mlcubes
- list of datasets
- benchmark detail
- mlcube detail
- dataset detail

# To run

```
medperf web-ui run [--port 8100]

# open http://localhost:8100/
```

# TODOs & bugs up to this moment

- common
  - [x] move date_format function to some common place and use everywhere
  - [ ] p1 build a unified render `mlcube-as-a-link` and use everywhere (benchmark's cubes, dataset's dataprep cube):
    - `[strong] mlcube name` `[small] modified date` `[state icon]` + colored in red if invalid?
  - [x] p1 hide `None` links
  - [x] p0 bug, no scrolling on main container (say, lists)
  - [x] p1 catch exceptions and display error
  - [ ] p2 dark theme
  - [x] p1 Bug, navigation is hidden when the page is too narrow
  - [ ] p3 use downloaded js packages (bootstrap etc.) instead of direct link?
- code
  - [x] p1 getting invalid entities raises an exception now
  - [x] p0 add templates & js to wheel package
- MLCube detailed page
  - [x] show tarball and additional data always
  - [ ] p2 check it works with local cubes
  - [ ] p3 add mlcube name to title
  - [x] p1 list of associated benchmarks + assoc status
- Dataset detailed page
  - [ ] p2 fix dataprep cube link render, see point higher ^^^
  - [x] p1 is_valid: make a badge? 
  - [ ] p2 Submitted as prepared: make a badge?
  - [ ] p3 what to do with `owner: None`? Show a hint over it describing why it's unknown?
  - [ ] p3 check it works with local datasets
  - [ ] p3 add dataset name to title
  - [x] p1 list of associations + status
  - [ ] p2 add metadatas
- Benchmark detailed page
  - [ ] p2 fix cube links render, see point higher ^^^
  - [ ] p2 add benchmark name to title
  - [x] p1 list of associated datasets + assoc status
  - [ ] p2 number of associated datasets
- MLCube list page
  - [ ] p2 redesign cards
  - [x] p1 list my entities on the top
  - [ ] p2 add valid filter
  - [ ] p2 add mine filter
  - [ ] p3 check it works with local cubes
- Bencmark list page
  - [ ] p2 redesign cards
  - [ ] p1 list my entities on the top
  - [ ] p2 add valid filter
  - [ ] p2 add mine filter
- Datasets list page
[...]

PR #617, opened 2024-09-16
[WIP] Web UI dataset

VukW:
depends on #606 
Adds functionality for creating & managing datasets

# Dataset submission

- [x] "add new dataset" button on datasets list
- [ ] "choose benchmark" step
  - [x] PoC (dropdown list)
  - [ ] `p1` list benchmarks as cards with detailed info (as on "benchmarks list" page), + highlight chosen one
  - [ ] `p3` allow to choose dataprep mlcube instead
- "Fill text fields" step
  - [x] DS name
  - [x] description
	- [ ] `p2` check length beforehand? DS submission would fail if description is > 20 characters ?
  - [x] location
    - [ ] `p2` check length beforehand?
  - [ ] `p3` "Submit as prepared" flag
    - needs to be explained somewhere (like, in a tooltip)
    - Q: if checked, is dataset automatically created as operational?
  - [ ] `p1` check how errors are displayed & handled
- "Paths" step
  - [ ] Q: combine with previous step?
  - [x] data path
  - [x] labels path
  - [ ] `p2` metadata path (what is it?) (is required if dataset is already prepared)
  - [ ] `p2` redesign path picking panel
    - [ ] folders / files difference
	- [ ] one click - chosing folder, double click - go inside?
  - [ ] `p2` "go back" button / navigation
- [x] "Verify entered data" step
- [x] dataset submission
- [ ] `p1` check how errors are displayed & handled

# Submitted dataset displaying (dataset details page)
- [ ] `p2` display paths (data path, labels path)
- [x] `p1` display dataset statistics

# Dataset preparation
- [x] "Prepare" button on dataset detail page
- [x] Preparation run
- [x] `p1` clean log messages from magic bytes
- [x] `p1` distinguish messages to headers + usual lines in the medperf code
- [x] `p1` display header messages properly
- [x] `p1` display log messages without json
- [x] `p2` log lines highlighting?
- [x] `p1` spinner at text headers to underline process is running
- [ ] `p1` check how errors are displayed & handled
- [ ] `p1` button "back to the dataset" rename
- [x] link to the report / display report if it exists
- [ ] `p1` check how errors are displayed & handled (display exceptions in the log)

# Prepared dataset displaying (dataset details page)

- [ ] `p2` if report exists, "Allowed automatic report submission" flag
- [x] `p2` if report exists, link / path to the report
- [x] `p1` if dataset is prepared, unlock next button "set operational" (locked if not prepared)

# Set operational

- [x] `p0` "Set operational" button on dataset detail page
- [x] `p0` Set operational
- [x] `p0` Disable button if already operational
- [x] `p1` check how errors are displayed & handled

# Associate

- [x] `p1` "Associate with the benchmark" button on the dataset detail page (no choice of benchmarks)
- [x] `p3` choice of benchmarks if dataset was created with dataprep mlcube
- [x] `p1` associate
- [x] `p1` check how errors are displayed & handled

# Run benchmark

- [x] `p1` "Run benchmark" button on the dataset detail page
- [x] `p1` running benchmark page with logs
- [x] ??? runs history??
- [x] ??? displaying result???
- [ ] `p1` check how errors are displayed & handled
- [x] `p2` "Run all" button
- [x] `p2` add "Running" status in addition to "Pending"
- [x] `p1` make all status (Run/Failed/Pending/Done) buttons as separate objects
- [x] `p1` modify default (right-after-load) button to "Waiting for status.."
- [ ] `p2` adjust logs panel height if there is no logs

# Submit result
- [x] `p1` "Submit result" button on the dataset detail page
- [x] `p1` Submit result
- [x] `p1` check how errors are displayed & handled
- [ ] `p2` "Submit all" button

# general dataset UI
- [ ] `p2` design buttons prepare-operational-... navigation to one line?
- [ ] `p1` hide buttons panel if you're not the dataset owner?
- [ ] `p3` redesign state displaying (we'd have dev/op floating blocks in the header + set-op button in the footer)

# Technical refactoring
- [x] `p1` split routes to different files: `dataset/submission.py`, `dataset/preparation.py`


cli/medperf/web_ui/api/routes.py
```diff
@@ -0,0 +1,33 @@
...
+
+BASE_DIR = os.getcwd()  # Restrict access to this base directory
+
+
+@router.get("/browse")
+def browse_directory(path: str = Query(...)):
+    full_path = os.path.join(BASE_DIR, path)
+    os.path.join(BASE_DIR, path)
+
+    if not os.path.exists(full_path) or not os.path.isdir(full_path):
```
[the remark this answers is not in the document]
VukW: This function lists all files and folders on the server side. So, yes, it is expected to use user's given path to list its content (though still may be dangerous? I don't know)

cli/medperf/commands/dataset/submit.py
```diff
@@ -43,11 +43,17 @@ def run(
...
+        submission_dict = preparation.prepare_dict(submit_as_prepared)
+        config.ui.print(dict_pretty_format(submission_dict))
+
+        msg = "Do you approve the registration of the presented data to MedPerf? [Y/n] "
+        warning = (
+            "Upon submission, your email address will be visible to the Data Preparation"
+            + " Owner for traceability and debugging purposes."
+        )
+        config.ui.print_warning(warning)
+        preparation.approved = preparation.approved or approval_prompt(msg)
```
aristizabal95, line 55: It would be better to instead convert this logic into a function, and keep the original flow with the added new function. Just so that we keep a consistent style where you can get an idea of the workflow of a command by looking at the `run` function.

cli/medperf/commands/result/create.py
```diff
@@ -1,4 +1,5 @@
 import os
+import time
```
aristizabal95, line 2: This dependency doesn't appear to be needed

cli/medperf/web_ui/api/routes.py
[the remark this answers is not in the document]
aristizabal95: We should handle all of these vulnerabilities as suggested, even if they're expected to be for local calls or not

cli/medperf/web_ui/datasets/routes_associate.py
```diff
@@ -0,0 +1,129 @@
...
+        dataset_id: int,
+        benchmark_id: int,
+        current_user: bool = Depends(get_current_user_ui),
+):
+    """
+    Serve the HTML page for associating a dataset with a benchmark.
+    """
+    dataset = Dataset.get(dataset_id)
+    benchmark = Benchmark.get(benchmark_id)
+    draft_id = str(uuid.uuid4())
```
aristizabal95, line 42: Didn't know of this. Nice!

aristizabal95 (approved):
This was a ton of work! There are a few suggestions on maintaining style, and covering vulnerabilities with the suggestions generated by github would be ideal. Also, this work should prompt some discussion on the current design and future design, since this implementation could not be added without repeating code. Leaving as is could lead to several bugs if changes need to be done in the current workflow (as they would need to be implemented in both CLI and Web UI). This is some amazing work for a task that was very intimidating and difficult to define

PR #618, opened 2024-09-17
Web UI profiles

VukW:
depends on #606, and #615
Web-UI: add a profile drop-down field to change profiles quickly


hasan7n:
closing as outdated

PR #624, opened 2024-10-29
Pass expected institutions during benchmark creation

aristizabal95:
This PR allows benchmark owners to optionally specify which sites are expected to participate on a benchmark. This is done through a CSV file that contains two columns: `email` and `institution`. This additiona provides the following benefits:

- Auto-approve users that are expected to join a given benchmark, reducing friction for participation
- Allow the benchmark owner to keep track of which institutions that are expected are currently participating or doing some work for the benchmark.

The institutions list are stored as a dictionary in the benchmark object. The CSV file is parsed by the medperf CLI to extract the information in the expected format. This list can only be visible by the benchmark owner.


hasan7n:
closing as outdated

PR #628, opened 2024-11-29
prevent empty statistics file

### LINKED ISSUE #623: [BUG] Empty statistics from a data owner causes a global error
**Describe the bug**
The `generated_metadata` field of datasets in the client is set to be required:
https://github.com/mlcommons/medperf/blob/09edc70bcf7f1615c532c8f8b15f10de9b406e33/cli/medperf/entities/dataset.py#L30
However, on the server, it is allowed to be `None`.
We faced a data owner whose generated statistics file happened to be empty, thus registering the `generated_metadata` field in the server as `null`. This caused a field validation error on the client side for any user that uses `medperf dataset ls`.

**Additional context**
We need to make the `generated_metadata` field optional on the client side

mhmdk0:
Fixes #623


aristizabal95:
Looks good! Amazing work @mhmdk0! My only suggestion if possible would be to use the mocked filesystem for testing, instead of patching `open` and `write`. This is because those operations are pretty common, and scattered throughout our codebase. If future changes will call those methods in other portions of the tested function then those patches would introduce side effects that will be difficult to figure out.

You can refer to this example:
https://github.com/mlcommons/medperf/blob/ea5a57636ac850c1767a36ebd646f58850961b27/cli/medperf/tests/entities/utils.py#L55-L70

As well as to the documentation of pyfakefs: https://pytest-pyfakefs.readthedocs.io/en/stable/usage.html#test-scenarios

If this is too cumbersome you can also leave it as is and create an issue as a reminder for later.

mhmdk0:
Just modified the tests to use fake fs instead of patching `builtins.open`. Thanks for informing me about this.

aristizabal95 (approved):
Awesome! Looks great!

Issue #622, opened 2024-10-28
[BUG] Dataset generated UID uniquness contraint fail in PostgreSQL


hasan7n:
**Describe the bug**
The constraint that no two operational datasets should exist with the same generated uid, seems to work well in development tests but fails in production.

https://github.com/mlcommons/medperf/blob/09edc70bcf7f1615c532c8f8b15f10de9b406e33/server/dataset/models.py#L39-L47

**To Reproduce**
Register two datasets in MedPerf with the same generated uid. Have one dataset in development and one dataset operational. Then, run `medperf dataset set_operational` on the development dataset. This returns 400 bad request when using the development version of medperf (as expected), but returns 500 internal server error using the production version of medperf.

**Expected behavior**
It should also return 400 bad request in production.
 
**Additional context**
This could be related to the fact that we use sqlite3 database in development, but we use postgresql in production.

PR #629, opened 2024-12-01
Fix generated_uid bug

mhmdk0:
Addresses #622 . Constraint failure causes _500 internal server error_ instead of _400 bad request_ for both **postgresql** and **sqlite3** .

`validate_state` function in [serializers.py](https://github.com/mlcommons/medperf/blob/eefdf16296000796c9066ebfd3947b3f480da698/server/dataset/serializers.py) wasn't working as expected as  [`self.instance.generated_uid`](https://github.com/mlcommons/medperf/blob/eefdf16296000796c9066ebfd3947b3f480da698/server/dataset/serializers.py#L44) isn't the new generated_uid coming from the put request body.

The code has been refactored and includes checks to prevent two **operational** datasets having the same **generated_uid** .


aristizabal95 (dismissed):
Looks good! My optional suggestion is that, since we're already revising this section of the code, it might be worth improving its style a little bit. I would suggest :
- splitting the logic into two functions to make it easier to understand the separation of concerns. One portion is dedicated to in-operation read-only fields, while the other is concerned with generated_uid collisions.
- Reducing nesting: The original check for the read-only fields has lots of nesting, which usually makes it a bit harder to read. There are some methods that remove nesting while keeping the logic equivalent. For example, the nested `if`s can be combined into a single `if` with an `and` operator. I personally strive to be a [never nester](https://www.youtube.com/watch?v=CFRhGnuXG-4) (even though I fail miserably some times)
- We can improve readability of the read-only checks by storing the conditionals in variables with descriptive names (e.g. `value_changed = v != getattr(self.instance, k)`).

I know that most of my suggestions are not directly related to the changes made, and come from already-existing code, but I thought it's a good chance to make some stylistic improvements. They're also completely optional! If you feel those changes are not needed feel free to merge as is 😄

mhmdk0:
Perfect! Just did the modifications, and it’s a great opportunity to make the code better overall as well.

Issue #625, opened 2024-11-04
[FEATURE] Have some logic for dataset export/import


aristizabal95:
**Is your feature request related to a problem? Please describe.**
We've noticed a common issue from real users related to moving from one machine to another. When this is done, users need to deal with our storage system, which can be cumbersome and error prone.

**Describe the solution you'd like**
Have something like `medperf dataset export <ID> [-o path/to/output]` and `medperf dataset import <ID>` to allow them to create a backup of the dataset and load it on another machine.

hasan7n:
closed by #632

PR #632, opened 2024-12-08
Implementing Import/Export Feature

mhmdk0:
Addresses #625.

Implements **import/export** feature that allow real users to move a dataset from one machine to another, handling all storage-related steps during the process.


hasan7n (approved):
Thanks Mohammad!

PR #644, opened 2025-04-23
Update MedPerf Documentation

### LINKED ISSUE #635: Documentation updates
Some portions of the current documentation page appear to be outdated or unclear. Here I propose changes to address issues I have found.

- [ ] Standardise documentation to use only `Data Owner` instead of both `Data Provider` and `Data Owner`. For example, [this section](https://docs.medperf.org/roles/#data-providers) uses `Data Owner` in the Figure, despite the section being named `Data Providers`.

- [ ] In the [Setup section for the local server](https://docs.medperf.org/getting_started/setup/#run-a-local-medperf-server), instruct users to use the virtual environment created in the [Installation section](https://docs.medperf.org/getting_started/installation/), if they made one.
  
- [ ] Add (conditional?) logout steps to the `setup_benchmark_tutorial.sh`, `setup_data_tutorial.sh`, `setup_model_tutorial.sh` and `simulate_data_association_approval.sh` scripts so that their login steps work properly. Currently, the logins fail if logged into a different user, which may happen for people doing multiple tutorials in a row.

- [ ] Fix the final `rm -rf` command in the [Clean Up](https://docs.medperf.org/getting_started/benchmark_owner_demo/#cleanup-optional) sections. The `~/.medperf/localhost_8000` directory mentioned does not exist. Rather, there are multiple `~/.medperf/<some_dir>/localhost_8000` directories to be removed instead. There is also a `~/.medperf/demo` directory that I believe is also related to tutorials and should also be removed.

- [ ] In [Section 2](https://docs.medperf.org/getting_started/benchmark_owner_demo/#2-develop-a-demo-dataset) of the Benchmark Committee section, explicitly mention that the workspace is the `medperf_tutorial` directory to avoid confusion, similarly to what is done in [Section 4](https://docs.medperf.org/getting_started/benchmark_owner_demo/#4-host-the-demo-dataset)

- [ ] In [Section 5](https://docs.medperf.org/getting_started/benchmark_owner_demo/#5-submitting-the-mlcubes) of the Benchmark Committee, the packaging script fails for the Data Preparation MLCube because the `mlcube.yaml` file does not have the `report_file` fields. It appears to be outdated.
    - This refers specifically to the command below. The packaging script works fine for the Model and Metrics MLCubes.
    ```bash
    python scripts/package-mlcube.py --mlcube medperf_tutorial/data_preparator/mlcube --mlcube-types data-preparator
    ```

    - If this error is ignored by the user, they can follow along the rest of the tutorial. The provided hosted MLCube (on DockerHub) is built correctly and compatible with the provided `mlcube.yaml` file. Therefore, the [submit step](https://docs.medperf.org/getting_started/benchmark_owner_demo/#data-preparator-mlcube_1) as presented in the current documentation works properly.

    - However, if the local `mlcube.yaml` file is updated to include the required `report_file` fields, the packaging script will work but submission will fail, as the image on DockerHub does not support a `report_file` input. So it would need to be updated too.

    - Another option: make the `report_file` field optional in the Schema.

RodriguesRBruno:
Closes #635 

I did not do changes regarding the packaging scripts due to MLCube support being dropped soon, so I imagine packaging will have bigger changes.

I also noticed a new MedPerf CA cube being introduced with a UID of 1 recently. This changes the UIDs of following cubes for the tutorials, so I updated these. I believe the UIDs will still be used even when transitioning to bare Docker.

Finally, I noticed the image hashes in `seed_utils.py` did not match the hashes of the corresponding docker images in DockerHub, which caused issues when running examples, so I updated the hashes in this file.


RodriguesRBruno:
I have reverted the hash changes to `seed_utils.py`. Seems that was  an issue on my end.

hasan7n:
closing as outdated

PR #649, opened 2025-06-27
Auto approve association

hasan7n:
Adds the feature of configuring the benchmark to auto-approve model/dataset associations either always or according to specified email addresses of the owners of the models/datasets.


cli/medperf/commands/benchmark/benchmark.py
```diff
@@ -224,3 +227,53 @@ def view(
...
     EntityView.run(entity_id, Benchmark, format, unregistered, mine, output)
+
+
+@app.command("update_associations_policy")
+@clean_except
+def update_associations_policy(
+    benchmark_uid: int = typer.Option(
+        ..., "--benchmark", "-b", help="UID of the desired benchmark"
+    ),
+    dataset_auto_approve_mode: str = typer.Option(
```
RodriguesRBruno, line 238: Typer supports Enums to both show the options on a standard format in their CLI help and also do validation on their side (including being case insensitive). See https://typer.tiangolo.com/tutorial/parameter-types/enum/ . I think this would be an improvement.

hasan7n: Thanks Bruno! That's indeed helpful

hasan7n: Thanks Bruno! That's indeed helpful

PR #657, opened 2025-07-22
Add Digests to sample data in seed_utils.py

RodriguesRBruno:
This PR adds the Digests to the sample data from `seed_utils.py`, so that the example cases can run properly on systems  (such as mine! :) ) where `docker inspect` returns the image digest rather than the hash we use.

I also took this opportunity to make the code download the `container_config.yaml` files from the main MedPerf repo (ie mlcommons/medperf; this repo!) rather than @hasan7n's fork.

### DIFF
--- a/server/seed_utils.py
+++ b/server/seed_utils.py
@@ -8,8 +8,8 @@ import django
 from django.contrib.auth import get_user_model
 
 ASSETS_URL = (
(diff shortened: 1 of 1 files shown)


hasan7n (approved):
Thanks Bruno!

PR #659, opened 2025-08-12
Add private model support

RodriguesRBruno:
This pull request adds support to the execution of Private Models in MedPerf.
The basic idea is that an encrypted file is used as the image (either an encrypted Docker Archive .tar or an encrypted .sif singularity image) and keeps an encryption key (secret) **LOCALLY** to decrypt this image. Once the model is registered to a benchmark, the model owner can encrypt the  is encrypted with Data Owners public keys (obtained from their public certificates) and the encrypted secrets are uploaded to the MedPerf server. When Data Owners run the benchmark, they download the encrypted secret and use their LOCAL private key to decrypt it, which allows access to the model


server/certificate/models.py
```diff
@@ -0,0 +1,22 @@
+from django.db import models
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+
+class Certificate(models.Model):
+    owner = models.ForeignKey(User, on_delete=models.PROTECT)
+    name = models.CharField(max_length=20, unique=True)
```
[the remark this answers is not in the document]
RodriguesRBruno: If we implement some sort of "medperf certificate ls" in the future, I think the name could be useful to distinguish them. I'm not sure if this would come up though.

I also implemented the name here (and on the EncryptedPrivateKey model too!) to keep it consistent with the client side pydantic models, as our MedPerfSchema base entity class has a name field. If there's a good way to omit the name while keeping this consistency, I think removing the name at least from the Key object would be a good idea! Can we do this?

server/certificate/permissions.py
```diff
@@ -0,0 +1,35 @@
...
+        if request.method != "GET":
+            return False
+
+        benchmark_id = view.kwargs.get("benchmark_id")
+        model_id = view.kwargs.get("model_id")
+
+        try:
+            benchmark_model_association = BenchmarkModel.objects.get(
+                model_mlcube__id=model_id, benchmark__id=benchmark_id
+            )
```
hasan7n, line 18: Not sure if `.get` here is correct to use. There could be multiple associations between a model and a benchmark (e.g., an approved one, then a rejected). That's how assocition rejection is handled in our server logic: the actual association between a benchmark and a model is the latest association object.

hasan7n: Check `IsAssociatedDatasetOwner` permission in `benchmark/permissions.py` file. The query is complex though and I have a TODO there to check for efficiency

server/certificate/views.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        certificate = self.get_object(pk)
+        serializer = CertificateSerializer(certificate)
+        return Response(serializer.data)
+
+
+class CertificatesFromBenchmark(GenericAPIView):
+    permission_classes = [IsAdmin | (IsModelApprovedInBenchmark & IsModelOwner)]
+
+    def get(self, request, benchmark_id, model_id, ca_id, format=None):
```
hasan7n, line 74: These two could be replaced by a `IsAssociatedModelOwner` similar to the `IsAssociatedDatasetOwner` permission from `benchmark/permissions.py` I think.

server/certificate/permissions.py
RodriguesRBruno: I modified the query to be similar to that is done in the ExperimentCA associations, which I think are more recent. Similar to what is shown below:
```
benchmark_model_association = BenchmarkModel.objects.filter(
            model_mlcube__id=model_id, benchmark__id=benchmark_id
        ).order_by('-created_at').first()
```

server/mlcube_ca_association/models.py
```diff
@@ -0,0 +1,55 @@
+from django.db import models
+from django.contrib.auth import get_user_model
+from django.http import Http404
+
+User = get_user_model()
+
+
+class ContainerCA(models.Model):
```
hasan7n, line 8: Let's discuss the following:
- to avoid having more entities and endpoints: let's treat this as if we were able to have a ManyToMany field in the mlcube object. (There will only be a POST endpoint here. Then we should edit the mlcube serializer to include a field representing the list of trusted CAs)
- It would be great to modify the training experiment CA association as well to become like this (it's unnecessarily complex the way we have it now. It shouldn't be an agreement between a CA and a training experiment; it should be just that a training experiment would or wouldn't like to trust a CA.)

RodriguesRBruno: Just pushed the changes about `ContainerCA` (didn't touch `ExperimentCA`). Now the `ContainerCA` model has been removed entirely and instead we have the `trusted_cas` ManyToMany field in the `MlCube` model.

server/mlcube_ca_encrypted_key/models.py
```diff
@@ -0,0 +1,30 @@
+from django.db import models
+from django.contrib.auth import get_user_model
+
+User = get_user_model()
+
+
+class ModelCAEncryptedKey(models.Model):
+    owner = models.ForeignKey(User, on_delete=models.PROTECT, null=True)  # Model owner
```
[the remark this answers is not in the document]
RodriguesRBruno: Included this in my last push. Makes things a lot simpler, thanks for the suggestion!

hasan7n (approved):

PR #660, opened 2025-08-19
Refactor Web-UI  Events, Tasks and Notifications

mhmdk0:
- Replace WebUI dictionaries with typed classes: `Events`, `Notification`, and `Task`.
- Use an event stream (SSE) for sending/receiving events.
- Add max log-event caps in both frontend and backend to limit stored logs and what’s shown in the panel, fixing memory spikes under heavy output.
- Introduce a new `Chunk` event that batches multiple log events with **size** and **age** limits to improve throughput and avoid flooding.

### DIFF
--- a/cli/medperf/config.py
+++ b/cli/medperf/config.py
@@ -231,6 +231,7 @@ logs_backup_count = 100
 cleanup = True
 ui = "CLI"
 webui = "WEBUI"
+MAXLOGMESSAGES = 200  # Maximum number of messages that will appear in LogPanel in WebUI
 
 default_profile_name = "default"
 testauth_profile_name = "testauth"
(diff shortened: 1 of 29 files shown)


mhmdk0:
**Commit [6a210d6](https://github.com/mlcommons/medperf/pull/660/commits/6a210d6f3375672494b20e0d0fa0c2d55ddb6941)**

- Fixed events streaming issue.
- Updated `medperf-login` to use the new events streaming logic and applied some front-end improvements.
- Fixed some front-end bugs in benchmark workflow test page.
- Fixed the issue preventing dataset owners from seeing the `Export Dataset` button.
- Corrected container compatibility test page redirection.
- Added counter badges for benchmark owners to view dataset and model associations, as well as results.

PR #693, opened 2026-06-07
Pyproject toml migration

mhmdk0:
This PR migrates MedPerf CLI packaging from cli/setup.py to a root-level pyproject.toml, so the project can be installed as a proper Python package (editable or from a wheel).

Main changes:

- Add root pyproject.toml and MANIFEST.in
- Remove cli/setup.py
- Keep package code in cli/medperf/
- Move mock_tokens/ into cli/medperf/mock_tokens/
- Fix tokens.json lookup to use the installed package path (fixes wheel installs)
- Update CI to use pip install -e ".[dev]"

Regarding OIDC configuration in PYPI, here's the official doc from github: [here](https://docs.github.com/en/actions/how-tos/secure-your-work/security-harden-deployments/oidc-in-pypi)


mhmdk0:
Added PyPI-based MedPerf update detection for the CLI and Web UI, with an optional one-click update from the browser (Web UI).

- UpdateManager — cached PyPI check, Web UI restart
- Web UI — update banner, manual “Check for updates”, “Update now” flow with progress overlay and auto-reload

mhmdk0:
## Summary

Migrates both `medperf` (CLI) and the Django server to pip-installable packages.

- **CLI**: `cli/pyproject.toml` packages `cli/medperf` as `medperf` (`pip install medperf`).
- **Server**: new `server/pyproject.toml` packages `server/` as `medperf-server` (`pip install medperf-server`), with a proper `medperf_server` CLI (`start`, `seed`, `reset_db`, `set_config <postgresql|sqlite|online-auth>`) replacing the old shell-script workflow (`setup-dev-server.sh`, `reset_db.sh`, manual `.env` copying).
- Renamed the Django project package `server/medperf` → `server/medperf_server` to avoid an import collision with the CLI's `medperf` package when both are installed in the same environment.
- Local dev config (mock auth keys/tokens, SSL cert, sqlite db, `.env`) now lives in a portable `~/.medperf_dev/` location instead of repo-relative paths, so it works consistently regardless of which directory commands are run from.
- Updated CI workflows and PyPI publishing (`publish-pypi.yml`) to match.

